PR #10, opened 2016-08-16
Follow standard node practice for disconnect callback

Pchelolo:
It's a standard practice in node.js to have error as the first argument in a callback. If we send and object as the first argument, automatic promisification (like in [bluebird](http://bluebirdjs.com/docs/api/promise.promisify.html)) of the `disconnect` method breaks.


webmakersteve:
Thanks for the PR! That was definitely an unforseen oversight because `disconnect` never really "errors"

PR #12, opened 2016-08-16
Fix Node 6 deprecation warnings

webmakersteve:


src/consumer.cc
```diff
@@ -394,7 +394,12 @@ v8::Local<v8::Object> Consumer::NewInstance(v8::Local<v8::Value> arg) {
 
   v8::Local<v8::Value> argv[argc] = { arg };
   v8::Local<v8::Function> cons = Nan::New<v8::Function>(constructor);
+  #if NODE_MAJOR_VERSION >= 6
+  v8::Local<v8::Object> instance = cons->NewInstance(Nan::GetCurrentContext(),
+    argc, argv).ToLocalChecked();
+  #else
   v8::Local<v8::Object> instance = cons->NewInstance(argc, argv);
+  #endif
```
Pchelolo: Better https://github.com/nodejs/nan/blob/master/doc/maybe_types.md#api_nan_new_instance

webmakersteve: Was looking for something like that in the docs but didn't see it. Thanks!

PR #13, opened 2016-08-16
Commit the message if it's set

Pchelolo:
I guess that's just a typo, but you were actually calling the function that's committing stored offsets when message was provided and function that's committing message when it actually wasn't provided. Or am I missing something here?


webmakersteve:
Yup! That was a typo :disappointed: . Thanks for catching it.

PR #14, opened 2016-08-17
Support building against librdkafka-dev package

Pchelolo:
In our setup we use docker to perform 'npm i' and to build all native dependencies in the same environment they will run in production. We maintain our own apt repository with `librdkafka@0.9.1` debian package, so we would like to build against the same version of the software as will run in production.

With this change I add a `BUILD_LIBRDKAFKA` env variable, defaulting to true, so in the default environment the behaviour of the build script would be unchanged. However, if you `export  BUILD_LIBRDKAFKA=0` before installing the package it will link to the librdkafka package installed on the machine. 

You could also benefit from this: if you install librdkafka locally and `node-gyp --BUILD_LIBRDKAFKA=0 rebuild` it will save you time on development since it will not recompile librdkafka sources every time.

In case you don't like this idea/change - feel free to close this PR, we will make our own fork and will be running it in production.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,29 +1,36 @@
 {
+  "variables": {
+      # may be redefined in command line on configuration stage
+      "BUILD_LIBRDKAFKA%": "<!(echo ${BUILD_LIBRDKAFKA:-1})",
+  },
   "targets": [
     {
       "target_name": "node-librdkafka",
-      "sources": [
-        "src/common.cc",
-        "src/errors.cc",
-        "src/config.cc",
-        "src/topic.cc",
-        "src/callbacks.cc",
-        "src/connection.cc",
-        "src/message.cc",
-        "src/consumer.cc",
-        "src/producer.cc",
-        "src/workers.cc",
-        "src/binding.cc"
-      ],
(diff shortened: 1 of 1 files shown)


webmakersteve:
Thanks for the PR. 

This is something I've played with in the past, but since we weren't using it like that I didn't include. However, I agree it is useful now that the library is open source.

webmakersteve:
I added CI for this in another branch:

https://travis-ci.org/Blizzard/node-rdkafka/builds/152860420

It's failing to link because it's looking for the `rdkafka++` library, naturally, but I can't see in the package file lists that that library is part of `librdkafka-dev`. Is there another package you're using that I can get on the Ubuntu CI server to test against?

Pchelolo:
They use a very old ubuntu distribution in travis (12.04) where librdkafka-dev isn't event packaged. Although even in trusty the the version is 0.8.3 (http://packages.ubuntu.com/trusty/librdkafka-dev) - it didn't include C++ api yet. 

You could probably some specific apt source to install the package from, but I've never done that (because of laziness and low priority), but that shouldn't be too hard to do..

PR #22, opened 2016-08-25
Update Default Partition Logic. 

rhysmccaig:
- Update default partition to -1 so that use of librdkafka's default partitioner is used when one is not specified (Currently defaults to partition 0.)
- Update documentation to indicate key and partition can be specified when producing a message

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,11 @@ producer.on('ready', function() {
     // converted to a Buffer automatically, but we're being
     // explicit here for the sake of example.
     message: new Buffer('Awesome message'),
-
+    // for keyed messages, we also specify the key - note that this field is optional
+    key: 'Stormwind',
+    // optionally we can manually specify a partition for the message
+    // this defaults to -1 - which will use librdkafka's default partitioner (consistent random for keyed messages, random for unkeyed messages)
+    partition: 0,
     // The topic object we created above
     topic: topic
   }, function(err) {
--- a/src/producer.cc
+++ b/src/producer.cc
@@ -415,7 +415,7 @@ ProducerMessage::ProducerMessage(v8::Local<v8::Object> obj, Topic * topic):
   m_is_empty(true) {
   // We have this bad boy now
 
-  m_partition = GetParameter<int64_t>(obj, "partition", 0);
+  m_partition = GetParameter<int64_t>(obj, "partition", -1);
 
   if (m_partition < 0) {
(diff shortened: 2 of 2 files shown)


rhysmccaig:
Pretty new to the contribution thing - let me know if i should do things differently.

src/producer.cc
```diff
@@ -415,7 +415,7 @@ ProducerMessage::ProducerMessage(v8::Local<v8::Object> obj, Topic * topic):
   m_is_empty(true) {
   // We have this bad boy now
 
-  m_partition = GetParameter<int64_t>(obj, "partition", 0);
+  m_partition = GetParameter<int64_t>(obj, "partition", -1);
```
Pchelolo: Perhaps better to use the constant: http://docs.confluent.io/2.0.0/clients/librdkafka/classRdKafka_1_1Topic.html#ab69c18779df1c8e150648e640b159654

rhysmccaig: Yeah i was just thinking that. Ill update.

rhysmccaig:
Updated to use the RdKafka::Topic::PARTITION_UA constant as the default instead of hardcoding -1.

rhysmccaig:
Sorry will fix that linting error.

README.md
```diff
@@ -164,7 +164,11 @@ producer.on('ready', function() {
     // converted to a Buffer automatically, but we're being
     // explicit here for the sake of example.
     message: new Buffer('Awesome message'),
-
+    // for keyed messages, we also specify the key - note that this field is optional
+    key: 'Stormwind',
```
webmakersteve: :heart:

webmakersteve:
Look's good to me. `-1` makes much more sense as a default partition - and was actually what I thought I was doing already. Thanks for catching this!

webmakersteve:
Merged. Thanks for the PR!

webmakersteve:
Just as an FYI - when submitting a PR I'd generally prefer the commits get squashed. You can do this using `git`'s interactive rebase.

GitHub now has this built in so it's super easy for me to do it on the fly during a merge, but it generally helps anyway :)

You can do this by running `git rebase -i HEAD~${NUMBER_OF_COMMITS}`. Then you can `p` commits to keep and `s` commits to merge. Then at the end you will rewrite the commit message. After you do that you can `-f` push to your branch and it will automatically update the PR without having to make a separate commit.

rhysmccaig:
Awesome - thanks for the guidance.

PR #24, opened 2016-08-26
Do null checking on partition instead of checking falsiness

webmakersteve:


rhysmccaig:
Nice work. Does the trick nicely.

webmakersteve:
Published.

rhysmccaig:
Wondering if the following 3 statements would make more sense (on their respective lines):

var gPart = conf.partition == null ? null : conf.partition;
this.defaultPartition = gPart == null ? null : gPart
var partition = msg.partition == null ? this.defaultPartition : msg.partition;


I think if the value is null (or maybe undefined?) in the c++ code then it should return the default code in the library? That way -1 isn't hardcoded in the javascript and in the c++ bindings.

(I don't really know c++ and node/v8 internals so I'm not sure if we are need null or undefined in order for the default to be set)

webmakersteve:
I definitely get the approach you're going for - we should try our best to not replicate constants for better maintainability. Unfortunately, some degree has to exist, and in this case it's probably better off we do defaults checking in node primarily, with c++ only as a fallback. The reason for the defaults in the getters is purely to work within C++ type safety and v8's unknown elements.

The real solution to this is making a C++ exported object with the constant properties, like `Topic::PARTITION_UA`. It's something I will get to eventually, but as it is unlikely to change the behavior of the app I'd rather put a bit on the backburner.

So in the case where I do properly export, it would be

`this.defaultPartition = gPart == null ? Kafka.CODES.TOPICS.PARTITION_UA : gPart;`

Then we can have the bridge we're looking for and the self-documentation of a constant for people who look at the Node code.

Issue #29, opened 2016-09-20
Assign / Unassign issue


ArtemisMucaj:
Hi, 

I've been trying to resolve an issue when adding consumer scripts with the rebalance callback.

1. Run X instances of the Consumer
2. Run another Consumer
3. Wait for rebalance and repeat 2.
4. From time to time one of the Consumers can't rebalance and it starts receiving `{"message":"Local: Erroneous state","code":-172,"errno":-172,"origin":"kafka"}` errors

This is what I get when running in gdb.

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7fe7780 (LWP 22887)]
0x00007ffff490b884 in NodeKafka::Consumer::NodeGetAssignments (info=...) at ../src/consumer.cc:493
493	      Nan::New<v8::String>(part->topic()).ToLocalChecked());

(I'm not entirely sure this is the reason)

ArtemisMucaj:
After further debugging I still don't really know what the core of the issue is (but I doubt its the segfault).
The one thing I keep observing is that sometimes "assign" method is not called and the Consumer can never recover.

ArtemisMucaj:
I figured out what causes the issue.

https://github.com/Blizzard/node-rdkafka/blob/master/src/consumer.cc#L156
https://github.com/Blizzard/node-rdkafka/blob/master/src/consumer.cc#L175

Changing `m_is_subscribed` state on `assign` and `unassign` doesn't make sense here as it causes `RdKafka::ERR__STATE` error when consuming even though `librdkafka` already maintains a `subscribed / unsubscribed` state internally.

Will write a PR asap.

webmakersteve:
Merged the PR so I believe the issue is closed!

As far as I can tell, it needs to be included in the next minor release because it can theoretically be a breaking change if someone relied on the throwing behavior in their application.

PR #30, opened 2016-09-20
Do not change subscribed state from assign / unassign methods (#29)

ArtemisMucaj:
Changing m_is_subscribed state on `assign` and `unassign` causes unrecoverable `RdKafka::ERR__STATE` error when consuming.

#29

### DIFF
--- a/src/consumer.cc
+++ b/src/consumer.cc
@@ -153,8 +153,6 @@ Baton Consumer::Assign(std::vector<RdKafka::TopicPartition*> partitions) {
   m_partition_cnt = partitions.size();
   m_partitions.swap(partitions);
 
-  m_is_subscribed = true;
-
   return Baton(RdKafka::ERR_NO_ERROR);
 }
 
@@ -171,8 +169,6 @@ Baton Consumer::Unassign() {
   if (errcode != RdKafka::ERR_NO_ERROR) {
     return Baton(errcode);
   }
-
-  m_is_subscribed = false;
 
   m_partitions.empty();
   m_partition_cnt = 0;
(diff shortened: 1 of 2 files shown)


webmakersteve:
The purpose of that variable is to ensure that when you consume, you are subscribed to a set of topics. This is the only way I can track that. How does this behave when you remove this piece?

ArtemisMucaj:
Removing it from `assign` and `unassign` methods fixes the issue. I don't think it's a good practice to modify your `m_subscribed` state here as unassigning / assigning doesn't mean that you're subscribed / unsubscribed to a topic! 

This is what happens :

1. reblance causes `unassign` to be called
2. `Consume` method will eventually throw `ERR__STATE` (https://github.com/Blizzard/node-rdkafka/blob/master/src/consumer.cc#L260) depending on wheter `RdKafka::KafkaConsumer` has queued a rebalance operation or not
3. If no operation was queued then we'll never call `assign` again and `m_subscribed` state will be stuck to `false` even though `librdkafka` is subscribed

It makes sense to track `subscribed`state but not to change it on `assign` / `unassign`.

ArtemisMucaj:
`subscribe` ->https://github.com/edenhill/librdkafka/blob/62bef22212ad5218bd08d6231184c49d36a2d98b/src/rdkafka_subscription.c#L52
`assign` -> https://github.com/edenhill/librdkafka/blob/62bef22212ad5218bd08d6231184c49d36a2d98b/src/rdkafka_subscription.c#L70

I just checked and they seem to be two different operation, maybe we should maintain two states instead of only having `m_is_subscribed`.

Whaddya think?

webmakersteve:
I think if we stop `assign` and `unassign` from modifying that variable and not relying on it to check the inner state before consumption (relying on `librdkafka` to do that for us) is the best solution. We can still use the value to determine whether the user has actually called subscribe to stop it from being called subsequently instead of failing silently. 

If you update the PR to do that I can merge it in after I test it.

ArtemisMucaj:
Will work on that asap.

Thanks

ArtemisMucaj:
@webmakersteve Does this seem good to you ?

PR #33, opened 2016-09-28
issue#32 - Consumed message has no key

### LINKED ISSUE #32: Consumed message has no key
Hi, we noticed that a message as received by a consumer has no key field.
Looking at the code, it seem not that complicated to add, we'll open a PR

edoardocomar:
added key field, using the std::string of rdkafka Message
codeveloped with @mimaison
fixes #32


edoardocomar:
oops will look into the segfault

src/message.cc
```diff
@@ -60,6 +62,8 @@ Message::Message(RdKafka::Message *message):
       size = message->len();
       offset = message->offset();
 
+      if (message->key())
+        key = *message->key();
```
webmakersteve: Can you add curly braces here? I think they're used in every if statement currently and would like to be consistent.

edoardocomar:
thanks @webmakersteve

PR #35, opened 2016-09-30
remove -lsasl2 from default build options in bindings.gyp

rhysmccaig:
Turns out that you can't build without sasl currently as -lsasl2 is always set in bindings.gyp, so i removed it.

As WITH_SASL=1 is the default condition and this sets -lsasl2 anyway this won't change the default behavior, but will allow WITH_SASL=0 builds to work.


webmakersteve:
Thanks for the PR! Nice catch!

PR #38, opened 2016-10-05
Support passing v8 regular expressions into subscribe

webmakersteve:
Currently one can use strings to pass regular expressions into subscribe but it requires knowledge of the wrapping syntax of `librdkafka`'s regular expression pattern. (i.e. `\c \"^${expression}\"`).

The update allows one to pass a v8 regular expression, e.g. `/^topic-.*/` or an array of regular expressions into subscribe and it will automatically convert it to the proper format.

### DIFF
--- a/lib/util/topicReadable.js
+++ b/lib/util/topicReadable.js
@@ -58,8 +58,8 @@ function TopicReadable(consumer, topics, options) {
   }
 
   if (!Array.isArray(topics)) {
-    if (typeof topics !== 'string') {
-      throw new TypeError('"topics" argument must be a string or an array');
+    if (typeof topics !== 'string' && !(topics instanceof RegExp)) {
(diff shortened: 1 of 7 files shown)


lib/util/topicReadable.js
```diff
@@ -58,8 +58,9 @@ function TopicReadable(consumer, topics, options) {
   }
 
   if (!Array.isArray(topics)) {
-    if (typeof topics !== 'string') {
-      throw new TypeError('"topics" argument must be a string or an array');
+    // Is there a better way to check if it is a regex?
```
InfinitiesLoop: Could probably remove this comment now :)

edenhill:
FYI: The next version of librdkafka (0.9.2) improves the wildcard subscriber to automatically pick up on newly created (matching) topics.

PR #39, opened 2016-10-09
Fix off by one error when commiting messages

InfinitiesLoop:
Note that the behavior of the underlying C bindings is correct, since the commit method exposed there wraps the vector-of-offsets commit method of librdkafka. Leaving that logic alone, but fixing the +1 here, leaves open the possibility of adding a commit overload to the javascript that lets users specify a specific commit offset without a corresponding message.


webmakersteve:
Going to redo this work and make it a separate method so we don't lose the ability to commit topic partitions.

Issue #46, opened 2016-10-21
Producer sends message with key length 0 even when key is present - only on Ubuntu 14.04 LTS


edoardocomar:
Hi we found a really mind boggling problem 

Using the following sample program, the key is sent with a length of 0 on Ubuntu 14.04.5 
(just retested on a clean VM!)
Note that **the code works on both Mac OS (10.11) and Ubuntu 16.04**

```
var Kafka = require('../');

var producer = new Kafka.Producer({
  //'debug' : 'all',  
  'metadata.broker.list': ...,
  'dr_cb': true
});

console.log('producer created');

//Connect to the broker manually
producer.connect();

var counter = 0;
producer.on('delivery-report', function(report) {
  // Report of delivery statistics here:
  console.log(report);
  counter++;
  if (counter === 10) producer.disconnect();
});

//Any errors we encounter, including connection errors
producer.on('error', function(err) {
  console.error('Error from producer');
  console.error(err);
});

//Wait for the ready event before proceeding
producer.on('ready', function() {
  console.log('producer ready');
  // Create a Topic object with any options our Producer
  // should use when writing to that topic.
  var topic = producer.Topic('testtopic', {
    // Make the Kafka broker acknowledge our message (optional)
    'request.required.acks': 1
  });

  console.log('topic created');

  for (var i = 0; i < 10; i++) {
    var message = new Buffer('message' +i);
    var key = "k-"+i;
    var partition = -1;
    producer.produce(topic, partition, message, key);
  }

  for (var j = 0; j < 50; j++) {
    setTimeout(function() {
      producer.poll();
    }, 1000);
  }
});
```

we could see that in `NAN_METHOD(Producer::NodeProduce)` in `producer.cpp`
if we add a debug after the key is created, around line 330:

```
    // This will just go out of scope and we don't send it anywhere,
    // since it is copied there is no need to delete it
    key = &keyString;
  }
 //add debug printf:
  printf("key->c_str=%s key->size=%d\n", key->c_str(), key->size());

  Producer* producer = ObjectWrap::Unwrap<Producer>(info.This());
```

then we can see that
`key->c_str()` will print the expected string ... but `key->size()` is 0 !!

sample output:

```
ubuntu@ubuntu:~/node-rdkafka/mickael-examples$ node producer2.js
producer created
producer ready
topic created
key->c_str=k-0 key->size=0
key->c_str=k-1 key->size=0
key->c_str=k-2 key->size=0
[...]

PR #53, opened 2016-11-01
attempted fix for key of zero length on Ubuntu 14

edoardocomar:
fixes issues #46 

tested on MacOS as well

developed with @mimaison

### DIFF
--- a/src/producer.cc
+++ b/src/producer.cc
@@ -330,7 +330,7 @@ NAN_METHOD(Producer::NodeProduce) {
   Baton b = producer->Produce(message_buffer_data, message_buffer_length,
     topic->toRDKafkaTopic(), partition, key);
 
-  //we can delete the key as librdkafka will take a copy of the message
+  // we can delete the key as librdkafka will take a copy of the message
   if (key) delete key;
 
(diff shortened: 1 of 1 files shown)


edoardocomar:
@webmakersteve we believe that this approach solves the bizarre issue we encountered on Ubuntu 14 - which is a platform that is essential for us.

src/producer.cc
```diff
@@ -322,18 +322,17 @@ NAN_METHOD(Producer::NodeProduce) {
...
+    key = new std::string(*keyUTF8);
   }
 
   Producer* producer = ObjectWrap::Unwrap<Producer>(info.This());
 
   Baton b = producer->Produce(message_buffer_data, message_buffer_length,
     topic->toRDKafkaTopic(), partition, key);
 
+  // we can delete the key as librdkafka will take a copy of the message
+  if (key) delete key;
```
webmakersteve: Can you add braces here to make sure it is a multi-line `if`.

e.g.

```
if (key) {
   delete key;
}
```

webmakersteve:
One change and I will merge.

webmakersteve:
Merged! Thanks for the PR.

PR #54, opened 2016-11-07
Don't point to the headers in the submodule explicitly

Pchelolo:
There're two options to build this repo controlled by the `BUILD_LIBRDKAFKA` parameter. If it's set to 1, the librdkafka sources are built from the submodule, and if it's 0, the `librdkafka-dev` package is used. In the latter case we need to point to the headers from the `librdkafka-dev` package as well, and not explicitly point to the cloned submodule. This is not only inconsistent, but it prevents us from pulling the dependency from github, since NPM doesn't do `git submodule update --init` when installing the deps, so even with `BUILD_LIBRDKAFKA=0` the build fails.

Now when we delete the explicit references to `deps/librdkafka/src-cpp/` exact location of the headers will be resolved by gyp `include_dirs` parameter.

Unfortunately, I've had to exclude the `rdkafkacpp.h` header from the cpplint check. The problem is that cpplint wants us to include a directory where the header is located, but with sources from submodule and librdkafka-dev the directory names would be different..

### DIFF
--- a/cpplint.py
+++ b/cpplint.py
@@ -399,8 +399,10 @@ _CPP_HEADERS = frozenset([
 # - Anything not following google file name conventions (containing an
 #   uppercase character, such as Python.h or nsStringAPI.h, for example).
 # - Lua headers.
+# - rdkafka.cpp header, because it would be located in different directories depending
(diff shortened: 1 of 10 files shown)


webmakersteve:
Thanks for the PR!

PR #56, opened 2016-11-14
Update librdkafka to 0.9.2

webmakersteve:


edoardocomar:
@webmakersteve Just FYI I've ran the whole build using the master branch e2e tests on my Mac with 0.9.2 and it had passed, 
including the test that with 0.9.1 we had left out see https://github.com/Blizzard/node-rdkafka/issues/63

webmakersteve:
Thanks for the info, @edoardocomar. Will test on my mac and see what the differences may be.

edoardocomar:
@webmakersteve would be great to have 0.9.2 
my users are screaming for getting seek/commited API on the js consumer

PR #60, opened 2016-11-15
Updated JSDoc for Producer#produce

Pchelolo:
The JSDoc should be updated after the Producer#produce was deprecated and proper way of producing is by providing `topic`, `partition`, `message` and `key` separately.


webmakersteve:
Can't believe I missed that. Thank you!

Pchelolo:
@webmakersteve Could you also update https://blizzard.github.io/node-rdkafka/current/ - a bit misleading..

webmakersteve:
They're updated. Need to get CI to do that on every build.

PR #62, opened 2016-11-16
Added a bunch of new e2e tests 

mimaison:
To help catch regressions early, we've added a bunch of new tests to the e2e suites. 

The assignment test in consumer.spec.js crashes once in a while:
```
*** ../deps/librdkafka/src/rdkafka_cgrp.c:790:rd_kafka_cgrp_partitions_fetch_start: assert: rkcg->rkcg_assigned_cnt <= (rkcg->rkcg_assignment ? rkcg->rkcg_assignment->cnt : 0) ***
Aborted (core dumped)
```
Is it because we're calling it incorrectly or is there a bug somewhere ?

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -76,7 +76,7 @@ describe('Consumer/Producer', function() {
     });
   });
 
-  it('should be able to produce and consume messages', function(done) {
+  it('should be able to produce and consume messages: subscribe/consumeOnce', function(done) {
     this.timeout(20000);
     var topic = 'test';
 
@@ -93,8 +93,7 @@ describe('Consumer/Producer', function() {
         offset = report.offset;
       });
 
-      consumer
-        .subscribe([topic]);
+      consumer.subscribe([topic]);
 
       var ct;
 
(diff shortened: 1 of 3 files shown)


webmakersteve:
Definitely appreciate adding more unit tests, but it looks like this one isn't passing.

There is an issue with mocha (which is why i shied away from it in the beginning), where if a test fails, it doesn't clean up after itself or force an exit. I'll test this PR locally and see what's going wrong so we can add the extra tests.

Those kind of core dumps are likely on the librdkafka side, but I think that assignment issue is a bug that was fixed in the most recent release of `librdkafka`. I am trying to get it to work, but there is currently a blocker. But when that gets merged in, it should fix the assignment issue.

edoardocomar:
We have found that on Ubuntu 14.04, if the producer test 
`' it('should produce a message to a Topic object' `
is run BEFORE  ` it('should get 100% deliverability',`
then the latter test hangs. 
That was the reason all 4 travis build failed.
It does not happen on MacOS. Changing the order should get the tests to pass in travis - we will need to open an issue for the producer hanging under those circumstances

edoardocomar:
temporarily excluded the `consumer.assign([])` test that may cause librdkafka to coredump

PR #72, opened 2016-12-07
Added message payload in the delivery report object

### LINKED ISSUE #70: Producer delivery reports - cannot figure out which message they relate to
The current delivery-report object contains key, offset, topic ... but this is not enough to figure out about which produced message they refer to.
Unless you wait for every single report after sending a single message.

Looking under the covers, node-rdkafka is using the full dr_cb_msg librdkafka callback, but does not put the message payload in the delivery report.

Investigated with @mimaison

edoardocomar:
changed the delivery report callback API, inserted err argument before
report

dr_msg_cb configuration used as flag to copy or not the msg payload in
the delivery report

message payload not copied on the error path (see callbacks.cc TODO)

fixes #70 
committed for discussion, cleanup (eg public flag in Dispatcher TODO)

developed with @mimaison

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ The library will throw an error if the value you send in is invalid.
 
 The library currently supports the following callbacks:
 * `partitioner_cb`
-* `dr_cb`
+* `dr_cb` or `dr_msg_cb`
 * `event_cb`
 
 ### SASL Support
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -88,8 +88,9 @@ describe('Consumer/Producer', function() {
 
       var offset;
 
-      producer.once('delivery-report', function(report) {
+      producer.once('delivery-report', function(err, report) {
         clearInterval(tt);
+        t.ifError(err);
         offset = report.offset;
       });
 
@@ -134,9 +135,10 @@ describe('Consumer/Producer', function() {
         producer.poll();
       }, 100);
 
-      producer.once('delivery-report', function(report) {
+      producer.once('delivery-report', function(err, report) {
         //console.log('delivery-report: ' + JSON.stringify(report));
         clearInterval(tt);
+        t.ifError(err);
         t.equal(topic, report.topic, 'invalid delivery-report topic');
         t.equal(key, report.key, 'invalid delivery-report key');
         t.ok(report.offset >= 0, 'invalid delivery-report offset');
--- a/e2e/producer.spec.js
+++ b/e2e/producer.spec.js
@@ -64,8 +64,9 @@ describe('Producer', function() {
       producer.poll();
     }, 200);
 
-    producer.once('delivery-report', function(report) {
+    producer.once('delivery-report', function(err, report) {
       clearInterval(tt);
+      t.ifError(err);
       t.ok(report !== undefined);
       t.ok(typeof report.topic === 'string');
       t.ok(typeof report.partition === 'number');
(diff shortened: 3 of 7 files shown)


edoardocomar:
@webmakersteve we agreed with your concerns about a 2nd memcpy of the payload,
so we propose to use the configuration option dr_msg_cb which looked like you thought of using for a while, to turn it into a flag to decide wether to do the memcpy or not.

This PR is still missing a bit of cleanup (the flag is a public member) and we haven't added the retrieval of the payload on the delivery report error path, which librdkafka provides

Also note the change in the delivery report callback API, from
`function(report)` to `function(err,report)`
and the fact that if there is an error, we emit both `delivery-report` and `error`.

rationale is that it's easier to have only one place,  `delivery-report`,  where to check and decide what to do (e.g. try and resend on failure, send another msg if ok)
but that `error` is anyway emitted for errors.

edoardocomar:
BTW, were you suggesting to use the *msg_opaque field as a client-generated message identifier to allow the client to determine which message failed? 
That appears to put a burden on the client code to store every message in a sort of map by ID to use on delivery-report.
May you expand a bit ? thanks!

edoardocomar:
@webmakersteve if you like this approach, we can finish up the TODOs in the implementation

webmakersteve:
@edoardocomar I was suggesting that we use `message_opaque` as an optional client-provided identifier for arbitrary data, as it is intended to be used I believe.

Example:

```js
const prod = new Kafka.Producer({}, {});

prod
  .connect()
  .on('ready', () => {
    prod.produce('topic', null, 'payload', null, { type: 'awesome' });
  });
```

Then we will be able to write this for our delivery-report

```js
prod
  .on('delivery-report', report => {
    if (report.msg_opaque.type === 'awesome') {
      console.log('We found the message');
    }
  });
```

webmakersteve:
I agree with changing this block so it passes the error in as the first parameter to the callback:

```js
    this._client.onDeliveryReport(function onDeliveryReport(err, report) {
      if (err) {
        self.emit('error', LibrdKafkaError.create(err));
      } else {
        self.outstandingMessages--;
        self.emit('delivery-report', err, report);
      }
    });
```

I think it makes sense to keep things in parity, 

The `dr_msg_cb`, I believe, was an older callback in `librdkafka`. I have mixed feelings about hijacking it to essentially act as some super-config. I would prefer the opaque method to fully solve the problem if you feel that will be adequate. If not, I think I would prefer opting in like this:

```js
const prod = new Kafka.Producer({ 'dr_cb': 'msg' });
```

Maybe that can be a decent enough way to specify you want the entire message in your delivery reports, including the payload.

Would definitely be open to thoughts here though.

edoardocomar:
@webmakersteve 
about the event, in your snippet above
```
    this._client.onDeliveryReport(function onDeliveryReport(err, report) {
      if (err) {
        self.emit('error', LibrdKafkaError.create(err));
      } else {
        self.outstandingMessages--;
        self.emit('delivery-report', err, report);
      }
    });
```
why having err in the else when it's always falsy ?

edoardocomar:
Hi @webmakersteve  about 
```
const prod = new Kafka.Producer({ 'dr_cb': 'msg' });
```
ok, but one could not register a specific function and have the payload. 

perhaps, looking again at 
https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
we could allow the user to set either `dr_cb` or `dr_msg_cb` and the difference would be the content. This looks to me to mimic librdkafka closer ?

@edenhill are you planning to keep both two settings ?

webmakersteve:
Was a typo from copy pasting. I just corrected it. And fair point, @edoardocomar. I figured most people used the event emitted interface instead.

edenhill:
@edoardocomar Both dr_cb and dr_msg_cb will remain, but using the latter is recommended since it provides a richer interface with more per-message information.
It is also future-proof in the sense that new message attributes will be available through dr_msg_cb, but not dr_cb.

edoardocomar:
Hi @webmakersteve  we've updated the PR ,
using dr_cb and dr_msg cb like librdkafka does. 

We've not implemented the msg_opaque suggestion, that could be done in addition to what we have.

Issue #68, opened 2016-11-29
producer acks property may not be honored


edoardocomar:
Even when creating a producer Topic object with code like this 
```
        // Create a topic object for the Producer to allow passing topic settings
        var topicOpts = { 'request.required.acks': -1 };
        var topic = producer.Topic(topicName, topicOpts);
        console.log('Topic object created with opts ' + JSON.stringify(topicOpts));

                producer.once('delivery-report', deliveryReportListener);

                var value = new Buffer('This is a test message #' + counter);
                var key = 'key';
                counter++;
                var partition = -1;

                    producer.produce(topic, partition, value, key);
                    // Keep calling poll() to get delivery reports
```
the produce request over the wire is sending acks=1 (the default).
We also tried setting the property value as a string` '-1'` or `'all'`. No joy.

Note that the property set in code is validated, i.e. an Error is thrown if the name or value are unknown. However valid values 0 and -1 are not honored.

We could verify this enabling tracing on the broker's kafkaApi 
sample log4j conf line:
```
log4j.logger.kafka.server.KafkaApis=TRACE, requestAppender
```
sample broker log:
```
[2016-11-29 17:53:10,103] TRACE [KafkaApi-0] Handling request:{api_key=0,api_version=1,correlation_id=114,client_id=mh-node-console-sample-producer} -- {acks=1,timeout=5000,topic_data=[{topic=mh-nodejs-console-sample-topic,data=[{partition=0,record_set=java.nio.HeapByteBuffer[pos=0 lim=56 cap=56]}]}]} from connection 192.168.1.67:9092-192.168.1.67:52624;securityProtocol:PLAINTEXT,principal:User:ANONYMOUS (kafka.server.KafkaApis)
```

edoardocomar:
Similar results using topic_opts in the Producer constructor and passing just the topicName to the produce method

webmakersteve:
Thanks for the report. I'll take a look at this in the C++ side and see if it's anything to do with node, or potentially a bug internally in `librdkafka`.

edoardocomar:
@webmakersteve FYI we tested with librdkafka 0.91 - using the C example included with the library - and there the setting is honored.

edoardocomar:
@webmakersteve we found the cause for the missing conf.

our code was calling 
```
        producer.getMetadata({
        topic: topicName,
            timeout: 10000
        },function(err, metadata) {...}
```
before calling 
```
[...]

PR #73, opened 2016-12-08
Document the side effect of client.getMetadata (#68)

edoardocomar:
Addresses issue #68 

investigated with @mimaison

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,9 @@ Getting metadata on any connection returns the following data structure:
 
 The following example illustrates how to use the `getMetadata` method.
 
+When fetching metadata for a specific topic, if a topic reference does not exist, one is created using the default config.
+Please see the documentation on `Client.getMetadata` if you want to set configuration parameters, e.g. `acks`, on a topic to produce messages to.
+
 ```js
 var opts = {
   topic: 'librdtesting-01',
--- a/lib/client.js
+++ b/lib/client.js
@@ -223,7 +223,7 @@ Client.prototype.connect = function(metadataOptions, cb) {
  * on the c++ wrapper you can do it here.
  *
  * @see connection.cc
- * @return {Conection} - The native Kafka client.
+ * @return {Connection} - The native Kafka client.
  */
 Client.prototype.getClient = function() {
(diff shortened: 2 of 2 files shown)


lib/client.js
```diff
@@ -304,6 +304,14 @@ Client.prototype.disconnect = function(cb) {
 
 /**
  * Get client metadata.
+ * <p/>
+ * <b>Note:</b> using a <code>metadataOptions.topic</code> parameter has a potential side-effect.
+ * A Topic object will be created, if it did not exists yet, with default options
```
webmakersteve: Typo: A topic object will be created, if it did not *exist* yet.

README.md
```diff
@@ -406,6 +406,8 @@ Getting metadata on any connection returns the following data structure:
 
 The following example illustrates how to use the `getMetadata` method.
 
+Please see the documentation for `Client.getMetadata` for a potential side-effect of the call.
```
webmakersteve: Can we reword this to something a little clearer. This seems a bit scary:

```
If you intend on passing a topic configuration to a topic you are explicitly fetching metadata for, be sure to created that topic prior to calling getMetadata on it. When fetching metadata for a topic, if a topic reference does not exist, one is created using the default config, which makes subsequent topic calls simply return that object.
```

We can and should also document it in the docblock as you have, but I think it's less cryptic if it is explicit in the README as well.

lib/client.js
webmakersteve: I have mixed feelings about using HTML markup in this comment when it isn't used in other places. At the very least, can we remove the `<p/>` and just leave the comment formatted as the other ones are?

webmakersteve:
Some comments on the PR and then I will be happy to merge.

lib/client.js
edoardocomar: ok, np!

README.md
edoardocomar: ok, made it a bit more explicit, less scary ;-) but still pointing to the docs. Duplication is evil ...

edoardocomar:
@webmakersteve changes ok?

PR #78, opened 2016-12-21
Create topics with the topic config, not default config

Pchelolo:
For #68 

When providing just a topic name to the producer the new `Topic` object is created, but it doesn't honour the topic config supplied to the producer, but instead creates a `Topic` object with the default config. This default config then takes precedence over the custom config making the driver ignore the custom config.

### DIFF
--- a/lib/producer.js
+++ b/lib/producer.js
@@ -111,7 +111,7 @@ function maybeTopic(name) {
     if (this.createdTopics[name]) {
       topic = this.createdTopics[name];
     } else {
-      topic = this.createdTopics[name] = this.Topic(name);
+      topic = this.createdTopics[name] = this.Topic(name, this.topicConfig);
     }
 
     return topic;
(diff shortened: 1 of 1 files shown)


webmakersteve:
It should be leveraging `librdkafka`'s default config, as it is set when the client is instantiated.

`producer.cc:37`
```
m_gconfig->set("default_topic_conf", m_tconfig, errstr);
```

We should not need to send it from JavaScript again. Are you seeing that it does not work? Perhaps it should be sending `null` or `undefined` in instead of an empty object.

Pchelolo:
Yes, I'm observing that this empty object overwrites the default config. See example in https://github.com/Blizzard/node-rdkafka/issues/68#issuecomment-268656288

Lemme try to use the `undefined`.

webmakersteve:
If that works, I'd prefer to leverage it that way. I am building a PR that should fix it to work like that.

webmakersteve:
It still may not work because it does checking on the parameters. If you could, see if this fixes your problem:

https://github.com/Blizzard/node-rdkafka/pull/81

webmakersteve:
Can you rebase this, and make it `topic = this.createdTopics[name] = this.Topic(name, this.topicConfig || undefined);`?

Pchelolo:
@webmakersteve done

PR #81, opened 2016-12-21
Send undefined in to leverage default config for producer.

webmakersteve:
Previously the producer was sending in a `{}` object and always creating a config. This will check if the parameter is set and only make a topic config if it is.

### DIFF
--- a/e2e/producer.spec.js
+++ b/e2e/producer.spec.js
@@ -97,7 +97,7 @@ describe('Producer', function() {
 
     producer.produce('test', null, new Buffer(''), '');
   });
-  
+
   it('should produce a message with a payload and key', function(done) {
     this.timeout(3000);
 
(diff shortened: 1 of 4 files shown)


Pchelolo:
Hm.. I've tried to do the same and although I can confirm that the `RdKafka::Topic` is created with the NULL config now and should fallback to default, it still doesn't work - the actual request has `RequiredAcks=1`

Pchelolo:
Even after this PR I see the same behaviour - the actual request to kafka has `RequiredAcks=1` while I'm setting it to -1. See for an example: https://github.com/Blizzard/node-rdkafka/issues/68#issuecomment-268656288

Pchelolo:
It seems like the `default_topic_config` is not honoured by librdkafka for the producer

webmakersteve:
That's pretty strange, as I would expect it to "throw" if it didn't take that config value. But I think we can take your PR in addition to make it work.

Pchelolo:
Strange indeed. Maybe a bug (not properly documented behaviour) of `librdkafka`?

PR #82, opened 2017-01-04
Add missing rdstring.c to librdkafka.gyp

0x7f:
When loading the latest native module based on the embedded deps/librdkafka, I get an error "symbol `rd_string_render` is missing". Turns out that `rdstring.c` (the file where `rd_string_render` is defined), is not included in the current node-gyp build script. This patch just adds the file.

Alternatively, the pull request https://github.com/Blizzard/node-rdkafka/pull/56 also adds it. But as long as it is not merged, it would be nice to have at least this patch merged.


webmakersteve:
Will merge this patch for the time being. There are conflicts in the 0.9.2 that need to be resolved anyway.

Issue #63, opened 2016-11-17
consumer.assign([]) may cause librdkafka to core dump


edoardocomar:
waiting for librdkafka 0.92
e2e test excluded temporarily

webmakersteve:
Should be fixed with #114

PR #86, opened 2017-01-12
Updated librdkafka to 0.9.2

mimaison:
We've grabbed the LZ4 changes from your pending PR and updated the librdkafka submodule link to tag v0.9.2.

All the tests (unit + e2e) pass on OSX but one of the producer e2e test fails on Ubuntu (14.04 and 16.04), see details below. Note that with librdkafka 0.9.1 all tests passed on both OSX and Ubuntu.

Work done with @edoardocomar 

The test that is failing is  `producer.spec.js - it should get 100% deliverability`:
```
Producer
    - should connect to Kafka
    - should produce a message with a null payload and null key
    - should produce a message with an empty payload and empty key (https://github.com/Blizzard/node-rdkafka/issues/36)
    - should produce a message with a payload and key
    1) should get 100% deliverability
    2) "after each" hook for "should get 100% deliverability"


  13 passing (55s)
  5 pending
  2 failing

  1) Producer should get 100% deliverability:
     Error: timeout of 3000ms exceeded. Ensure the done() callback is being called in this test.
  

  2) Producer "after each" hook for "should get 100% deliverability":
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

Reducing the number of messages to a few hundreds (~300) makes the test pass again. We've enabled librdkafka debug logging and when it's failing we see these 2 messages keep being printed forever:
```
{ severity: 7,
  fac: 'TOPPAR',
  message: 'mickael-VirtualBox:9092/0: test [0] 0+0 msgs' }
{ severity: 7,
  fac: 'SEND',
  message: 'mickael-VirtualBox:9092/0: Sent partial ProduceRequest (v1, 0+0/378942 bytes, CorrId 2)' }
```

and the broker does not receive any of the messages.

Attached is the full output (I've only kept a few of the looping messages, but we've let it run for about 10 mins and it was still printing them over and over)

[output.txt](https://github.com/Blizzard/node-rdkafka/files/702390/output.txt)

cc @edenhill


edoardocomar:
interestingly, the issue goes away switching to the current (commit 6745465) trunk of librdkafka .
But I'm uneasy on working off trunk of a dependent project.
There must have been a fix to librdkafka between 0.92 and trunk that is relevant for node-rdkafka.
@edenhill ?

edenhill:
@edoardocomar I dont find anything obvious in the commits since v0.9.2.
If the tests are automated you could give 'git bisect' a try to let it find the fixing commit.

webmakersteve:
https://github.com/Blizzard/node-rdkafka/pull/56 has the exact same issue which is why it hasn't been merged. If you run the tests with `DEBUG=all` you will see some strange messages sent on the producer side. 

I'm trying very hard to get to debugging the issue, but it's the first full week after our break. It's my first priority once things stabilize a bit.

mimaison:
We've had a go with bisect and it looks like since v0.9.2 this test got broken and fixed several times.
For example, 100% deliverability (librdkafka commits in chronological order):
- fails with v0.9.2
- passes with 0b0ccecf3a2bebd6338beb1cca98a198f4d52d10
- fails with a7346fc734b454c56b4c9bb61be58117ae85f2db
- passes with 4e4fae10d3a9b1b582de6819fa69ab724181b3a6

Even with this test passing, at many of these commits, other tests would fail.

We get the best results with the current librdkafka HEAD (bbee686). Using it, all tests pass on Linux and OSX. And also it fixes #63 !

webmakersteve:
I'm a little nervous to merge a branch that points to a commit not officially marked as a release. Did you run the tests more than once to make sure it wasn't just a fluke?

edenhill:
There will be a 0.9.3 release within a couple of weeks with some hotfixes since 0.9.2.
A larger 0.9.4 release will be available later.

Maybe you should target the `0.9.3.x` branch?

mimaison:
@webmakersteve Yes we should only use librdkafka releases. I only updated the PR to verify librdkafka master was passing the e2e tests in Travis.

With 0.9.3 round the corner, let's wait for it.
We've tested the current librdkafka 0.9.3.x and all tests pass on Linux and Mac

mimaison:
Superseeded by #88

PR #87, opened 2017-01-13
Simplified librdkafka.gyp

edoardocomar:
using find with wildcards and exclusions, instead of explicit file list

### DIFF
--- a/deps/librdkafka.gyp
+++ b/deps/librdkafka.gyp
@@ -14,17 +14,7 @@
         "librdkafka"
       ],
       'sources': [
-        "librdkafka/src-cpp/RdKafka.cpp",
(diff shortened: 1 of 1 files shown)


deps/librdkafka.gyp
```diff
@@ -44,13 +34,13 @@
         ['OS=="mac"', {
           'xcode_settings': {
             'OTHER_CFLAGS': [
-						  '-ObjC'
-					  ],
+                          '-ObjC'
```
webmakersteve: Can you fix the indentation here and below in the `-stdlib...`

webmakersteve:
There are some code formatting issues that are messing up the diffs a bit. Once fixed I can merge. You will need to rebase the other PR if we do though.

PR #90, opened 2017-01-16
In ConsumeLoop, only subscribe if there is no existing assignment/sub…

### LINKED ISSUE #89: can't use consumeLoop with assignments
In `kafka-consumer.js` `consume([topics])` calls `subscribe([topics])`
therefore client code like this
```
  consumer.assign([{ topic:topicName, partition:0, offset:2950}]);
  consumer.consume([ topicName ]); 
```

does not work, because the assignment (i.e. unbalanced consumer with manual partition assignment) is overridden by the hidden subscribe (i.e. causing a shift to the balanced consumer)

the consumer API should allow both these use cases:
```
  consumer.assign([{ topic:topicName, partition:0, offset:2950}]);
  consumer.consume();  //invoke the loop
```
and

```
  consumer.subscribe([topics]);
  consumer.consume(); //invoke the loop
```
The consume method should not require again the list of topics that have been subscribed or assigned to.

This behavior looks more consistent with the Java client, though removing the argument from `consume` currently invokes `consumeOne`

mimaison:
…scription

Fixes #89

Work done with @edoardocomar

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -74,8 +74,7 @@ describe('Consumer/Producer', function() {
     });
   });
 
-  it('should be able to produce and consume messages: subscribe/consumeOnce', function(done) {
-    this.timeout(20000);
+  function testConsumeOne(done, doSubscribe) {
     var topic = 'test';
 
     crypto.randomBytes(4096, function(ex, buffer) {
@@ -84,15 +83,16 @@ describe('Consumer/Producer', function() {
         producer.poll();
       }, 100);
 
-      var offset;
-
       producer.once('delivery-report', function(err, report) {
         clearInterval(tt);
         t.ifError(err);
-        offset = report.offset;
       });
 
-      consumer.subscribe([topic]);
+      if (doSubscribe) {
+        consumer.subscribe([topic]);
+      } else {
+        consumer.assign([{ topic: 'test', partition:0 }]);
+      }
 
       var ct;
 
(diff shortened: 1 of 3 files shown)


edoardocomar:
@webmakersteve as mentioned in https://github.com/Blizzard/node-rdkafka/issues/89 I'd like not to have to pass an array of topics in `consume()` and still have the consume loop behavior.  

In this PR we've tried not to break the existing API (too much).

Would you consider instead breaking the API or deprecating the current consume and add a publicly visible `consumeLoop()` with no arg ?

webmakersteve:
I think what you're describing is the better solution here. I don't really like doing checking on the client to see whether to call subscribe. The original goal with consumeLoop is to consume off a topic automatically with the higher level consumer.

I'll add support for consuming without the automatic behavior in some other form, using a different API.

edoardocomar:
Hi @webmakersteve not sure what you're suggesting we do:
- add a public `consumeLoop()` method 
or
- have the current `consume()` invoked with no args invoke the current, private, `consumeLoop`
having as pre-req that `assign([tparts])` or `subscribe([topics])` are called explicitly before `consume()` ?
or
- ...

please could you suggest. 
We need this fix in so that, with the addition of rebalance events we can use assign with offset to have the behavior of seeking to an offset ... whether the consumer is using subscriptions or direct assignment.

webmakersteve:
Sorry. Wasn't clear.

What I'd like to do is make it so consume can take `null` as the first set of parameters to specify not to subscribe, so it can be done manually. So rather than calling it with a topic list and a callback, you would call it with null and a callback to tell it you still want the consumeLoop behavior, but you do not want it to automatically do assignments/subscription.

Additionally, when using the Readable stream, I'd like to refactor that separately so you provide the consumer rather than have it be a method on the consumer. There may be some initialization options you can pass it to automatically subscribe, but I'd like to make the stream a first class citizen.

webmakersteve:
I've added an implementation in #96. Does this work for you?

webmakersteve:
This fix is made unecessary with the new consume changes I believe. Please feel free to reopen if I'm wrong! #103

PR #94, opened 2017-01-17
MINOR: gitignore Eclipse generated files

edoardocomar:


mattness:
I'm generally -1 on this one.  IDE-specific ignores belong in the gitignore of the user of the IDE.  Adding eclipse here opens the door to having to add every other IDE that a user might use.  At that point, the project's ignore list becomes difficult to maintain because it's all noise from IDE ignores instead of signal from the project.

You can create a user-specific .gitignore (which will get merged at runtime with each repo's gitignore) by running the following:

```bash
git config --global core.excludesfile $HOME/.gitignore
```

Where `$HOME/.gitignore` can be any file you like.  See [GitHub's help docs](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for more info.

edoardocomar:
Thanks @mattness 
Closing

PR #96, opened 2017-01-18
Allow null to be passed into consume for topics

webmakersteve:


mimaison:
Thanks @webmakersteve for suggesting a change.  At this stage it feels like we're trying too much to not break the API. Consume() is really overloaded and this new option does not feel natural at all. 

As @edoardocomar hinted in https://github.com/Blizzard/node-rdkafka/issues/89#issue-201074691, we're suggesting changes to clean up the consume API and make it easy to use and flexible.

Suggested new API:
Consume now requires `assign([...])` or `subscribe([...])` to have been called.

- `consume(callback):` optional callback argument:
It works like current `_consumeLoop` but does not do automatic `subscribe()`.

- `consume(number, callback)`:  required number argument, optional callback argument:
It consumes `number` messages. It works like current `_consumeNum` (can be called with number = 1 to replicate `_consumeOne`).

If we want to make migration easy, we could just deprecate the current `consume(...)` and add this new API under a different name, like `poll(...)` for example.

webmakersteve:
I agree with the changes you've put forward. I do think consume is too overloaded, and the changes you've suggested make sense. We're still in a sub 1 version number so I think making breaking api changes is especially okay as long as we bump the version up by at least `0.1`, which has been the pattern so far.

PR #101, opened 2017-01-19
Improvements to the delivery report using the opaque (issue#100).

### LINKED ISSUE #100: Improvements to the delivery report using the opaque
The addition of the opaque is great (performance/flexibility) but:

1. In order to have the value in the delivery report, it forces the user to make a "clunky" API call.
For example:
```
var value = new Buffer('...');
var key = 'key';
producer.produce('topic', null, value, key, {value: value, key: key});
```

It feels a lot more natural to automatically get the opaque populated in case the user has not passed in a custom value. And this does not cause any performance penalty as it reuses the opaque mechanism.

2. The key member of the delivery report object still suffers from the null/empty issue. It can be correctly retrieve for free from the opaque as described above, hence removing the need to have native code handling the key. 

Although people generally consider to be small, in Kafka it is actually the same datatype as the payload and can be arbitrary big. So reusing the opaque to pass it along the payload make sense.

edoardocomar:
Fixes #100

Automatically generated opaque.
Removed delivery report key handling in native code.
Added tests.
Removed failing test for null key in delivery report.

Developed with @mimaison

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -144,7 +144,6 @@ describe('Consumer/Producer', function() {
         clearInterval(tt);
         t.ifError(err);
         t.equal(topic, report.topic, 'invalid delivery-report topic');
-        t.equal(key, report.key, 'invalid delivery-report key');
         t.ok(report.offset >= 0, 'invalid delivery-report offset');
       });
 
--- a/e2e/groups.spec.js
+++ b/e2e/groups.spec.js
@@ -14,7 +14,7 @@ var Kafka = require('../');
 var kafkaBrokerList = process.env.KAFKA_HOST || 'localhost:9092';
 var eventListener = require('./listener');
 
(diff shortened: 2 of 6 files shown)


edoardocomar:
updated the PR - automatic opaque only created with `conf.dr_msg_cb` not with `conf.dr_cb`

please @webmakersteve review

webmakersteve:
I added the `dr_msg_cb` back in as I found it was the wrong call thinking the opaque could solve all use cases. Does this solve the issue so we can close this PR? Or do you find that there is more

edoardocomar:
Thanks, we will take a look in a few days

mimaison:
@webmakersteve Sorry for the delay. I had a quick look today and yes it looks fine. Thanks!

edoardocomar:
this PR is now obsolete . 
Thanks @webmakersteve and  @mimaison

PR #103, opened 2017-01-19
Update consume method api to be less 'magic'

webmakersteve:
Leaving this PR here for discussion as development is made on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,10 +316,12 @@ consumer.connect();
 
 consumer
   .on('ready', function() {
+    consumer.subscribe('librdtesting-01');
+
     // Consume from the librdtesting-01 topic. This is what determines
-    // the mode we are running in. By consuming an entire topic,
-    // we will get messages from that topic as soon as they are available
-    consumer.consume('librdtesting-01');
+    // the mode we are running in. By not specifying a callback (or specifying
+    // only a callback) we get messages as soon as they are available.
+    consumer.consume();
   })
   .on('data', function(data) {
     // Output the actual message contents
@@ -339,7 +341,7 @@ consumer
 
     // Read one message every 1000 seconds
     setInterval(function() {
-      consumer.consume();
+      consumer.consume(1);
     }, 1000);
   })
   .on('data', function(data) {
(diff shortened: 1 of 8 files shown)


edoardocomar:
Hi @webmakersteve we like the API proposal :-) great!

Note that you have dead code now in the binding, e.g.
https://github.com/Blizzard/node-rdkafka/blob/consume-changes/src/consumer.cc#L834-L845
https://github.com/Blizzard/node-rdkafka/blob/master/src/workers.cc#L424-L443

as you now will always use ConsumeNum the callback to consume will always pass an array `[messages]` no longer a single `message` - which is more consistent.

Also, the condition 
https://github.com/Blizzard/node-rdkafka/blob/master/src/workers.cc#L436-L437
is different from 
https://github.com/Blizzard/node-rdkafka/blob/master/src/workers.cc#L360-L361

that's why the ifs in this block are never true
https://github.com/Blizzard/node-rdkafka/blob/master/e2e/both.spec.js#L101-L108
and the test fail.

It can be made to pass by checking for the empty messages array
```
          if (messages.length === 0) {
            producer.produce(topic, null, buffer, null);
            ct = setTimeout(consumeOne, 1000);
            return;
          }
```

edoardocomar:
@mimaison and me discussed these methods:
https://github.com/Blizzard/node-rdkafka/blob/master/src/workers.cc#L354-L369
https://github.com/Blizzard/node-rdkafka/blob/master/src/workers.cc#L396-L411

And we're wondering if a bug could be hidden in the `consumeNum` logic.
If after successfully doing N<max iterations inside `ConsumerConsumeNum::Execute()`, 
the subsequent call to `m_consumer->Consume(m_timeout_ms);` returns an error, 

Then `ConsumerConsumeNum::HandleErrorCallback()` seems to delete the N messages already fetched and return only an error up the stack.

Are those first N messages lost now ? 
Isn't `librdkafka` assuming those have been fetched and so the consumer position (and possibly committed offsets) advances without those messages having actually be seen by the application ?

Note that this behavior is unrelated to the new proposed JS consumer API. It is existing behavior,

webmakersteve:
I have updated the ConsumeNum logic to only bubble the error up to the worker if it is the first iteration of the loop. 

Additionally I implemented the fix for the test.

Issue #93, opened 2017-01-17
Consumer issues with `conf.rebalance_cb = true`


edoardocomar:
if `conf.rebalance_cb` is set to `true` 
the consumer does not emit the rebalance event - which could be used to mimick `seek` by using `assign` as suggested by @edenhill

BTW with conf.rebalance_cb = true`, the consumer calls an undefined `unassign` JS function

`kafka-consumer.js` line 54
```
        self.unassign(e.assignment);
```

webmakersteve:
Should be fixed with #95.

edoardocomar:
Thanks @webmakersteve !

Will need to add some tests. 
I remember seeing `unassign` throwing because of client disconnected. will need to retest that path too.

Would be nice to be able to mark the the event as "manually handled" and if so, skip the following `if...assign...else...unassign`

webmakersteve:
You can do that already by providing a function into `conf.rebalance_cb`

edoardocomar:
I was thinking, as the "standard" rebalance action is already coded in the `KafkaConsumer`
```
      if (e.code === 500 /*CODES.REBALANCE.PARTITION_ASSIGNMENT*/) {
        self.assign(e.assignment);
      } else if (e.code === 501 /*CODES.REBALANCE.PARTITION_UNASSIGNMENT*/) {
        self.unassign();
      }
```
that could be extracted in a method that can be called in the user's event handler (eg after changing the offset in `e`) , and then skipping the re-invocation of assign/unassign

webmakersteve:
The way I envisioned it, is that the `rebalance` event being emitted is just extra information we are aware of if we opt into our software handling the rebalances. If you want to override the functionality, it should be done inside a function passed directly into the config, rather than in the emitter. This seems to me the most 1:1 relationship to `librdkafka`, while still adding some extras that make sense.

So, if you want `node-rdkafka` to manage rebalances, but want to be notified, you...

```js
const consumer = new kafka.Consumer({
   'rebalance_cb': true
});
```

If you want to implement custom rebalance logic...

```js
const consumer = new kafka.Consumer({
   'rebalance_cb': function(event, assignments) {
      this.assign(assignments); // this is bound to the kafka.Consumer via bind
   }
});
```

If you'd like to see it a different way, one of the ways I was thinking is to do it the way you've described. But so far this is the API i've come up with.

edoardocomar:
I have written a test that shows the problem I mentioned in
[...]

PR #104, opened 2017-01-20
WIP For discussion - Issue #93 when conf.rebalance_cb is set to true

edoardocomar:
consumer.disconnect() causes the self.unassign() in the rebalance
function to be invoked when the consumer is in the wrong state.

the test `should emit events with conf.rebalance_cb=true`
passes but the testsuite hangs (process does not exit)


webmakersteve:
I think this was fixed in a past release, so I am closing the PR.

PR #116, opened 2017-02-09
MINOR: e2e/consumer empty assignment test neabled

edoardocomar:
it passes with librdkafka 0.93


edoardocomar:
developed with @mimaison

Issue #117, opened 2017-02-09
null/empty key/value not honored in delivery report


mimaison:
When sending a message with an empty key, the delivery report we receive has a `null` key instead of an empty string. 
With dr_msg_cb, when sending an empty payload, the delivery report we receive has an `undefined` payload

webmakersteve:
Fixed with commit ef49abe651ab02c46128b3284a03ed93bbe5daaa. Will get pushed with update to 0.8

PR #118, opened 2017-02-09
null/empty key/value not honored in delivery report

edoardocomar:
testcases for issue #117 

developed with @mimaison


webmakersteve:
Test is currently failing, which I assume is the purpose? I'll bring these local and fix the issue

edoardocomar:
Yes @webmakersteve, the failing test is to prove and reproduce the issue - and assume to be merged with the fix

Issue #115, opened 2017-02-09
Tests not passing


jmarca:
Hi, just learning kafka etc., so it is likely I'm doing something wrong.

I followed the instructions for running tests.  I cloned this repo, did `git submodule init` followed by `git submodule update`, then did an `npm install` to download dependencies.  

Then, in my kafka directory, I followed the "getting started guide" and fired up zookeeper and kafka as follows:

In screen 1, I did 

``` 
bin/zookeeper-server-start.sh config/zookeeper.properties
```
and in another screen session I did 
```
bin/kafka-server-start.sh config/server.properties
```
(Note both of these used the default server.properties that are distributed with Kafka source code, version 2.11-0.10.1.0) (in other words, brain dead clueless settings)

Then to run the tests I ran `make`.  The first time through I got two errors.  Then I ran make again and just got one error (which makes me suspect the first, non-repeating error happened because something wasn't right in kafka the first time through, then was okay the second time??)

The failing test says:

```
  1) Producer with dr_cb should connect to Kafka:
     Uncaught AssertionError: Broker id is not set
      at e2e/producer.spec.js:53:11
      at lib/client.js:351:7
```

Running on slackware 14.2 (-current), node v7.2.0

webmakersteve:
Do you have any way of checking the metadata returned by your Kafka broker? You can use KafkaCat, which uses `librdkafka` under the hood just like `node` does.

```sh
kafkacat -L -b broker:9092 [-t topic]
```

Note: topic is optional

Just see what `librdkafka` is returning back for your set up. There is a chance that the broker ID isn't set so `librdkafka` doesn't know what to do with the metadata it is being given.

jmarca:
Haven't had a chance to get back to this, but I didn't want to mention a few items I noticed between posting and getting your reply. (I did just try to quickly npm install -g kafkacat but no executable got installed, so I'll return to that later)

I was able to make the tests pass once.  I started a cluster of three "nodes" on my laptop.  I messed up and started in the order node 1, node 2, and then node 0 on the default port.  Then the tests passed.  No idea why, I wasn't able to reproduce that starting the three nodes in order of 0, 1, 2.

Second, I get the "double error" if I delete the temporary zookeeper and kafka directories in the /tmp directory (again, I'm following the quickstart guide on the kafka site)

[...]

PR #121, opened 2017-02-15
test orig_broker_id vs undefined

jmarca:
if orig_broker_id is 0, then t.ok(0) will fail.

change test to t.ok(0 !== undefined) the test will pass

fixes issue #115


webmakersteve:
Thanks for the PR!

PR #130, opened 2017-03-02
Closing the TopicReadable calls the callback

sgenoud:
The callback is called on the `'close'` event – while the code emitted a `'closed'` one. I chose to stick with the former one, as it is what standard readables emit.


webmakersteve:
Merged. Thanks for the PR.

PR #134, opened 2017-03-03
commitSync throws errors as they happen

sgenoud:
It used to return error objects instead.

This seems to me the better API – but might be breaking some code. Personally I was surprised by this behavior and had a bug that assumed that because the commitSync did not throw.

@webmakersteve there are a couple of other issues with commit / commitSync that I want to look at / work on, would you prefer separate PRs?

### DIFF
--- a/e2e/consumer.spec.js
+++ b/e2e/consumer.spec.js
@@ -28,6 +28,36 @@ describe('Consumer', function() {
     };
   });
 
+  describe('commit and commitSync', function() {
+    var consumer;
+    beforeEach(function(done) {
+      consumer = new KafkaConsumer(gcfg, {});
+
+      consumer.connect({ timeout: 2000 }, function(err, info) {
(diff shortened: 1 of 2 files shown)


webmakersteve:
This should absolutely throw. What version did you see this behavior change? It was definitely unintended that it ever returned error objects. 

Also, I would prefer separate PRs. Make sure you rebase your code with the current master as I just merged a commit method change.

sgenoud:
I have not seen the behaviour change (but the eldest version I used was 0.7.0-alpha so it is not very interesting).

In my rebase I also change the return value to follow your documentation.

webmakersteve:
Sorry. Was making some other changes to commit, but definitely want to give you credit! Can you do this for the other methods, and then I will merge! :)

sgenoud:
Merge this, I was busy with other things and did not do any futher work (yet?)

PR #137, opened 2017-03-07
Tests for commit behavior

thanpolas:
Resubmitting this to add the commit tests. I am afraid they will not play nice with other tests as they leave a `consumer.once()` untriggered... if you think that's ok let me know.

I'll push the changes to meet the new API for commitMessage() in a few

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -247,7 +247,7 @@ describe('Consumer/Producer', function() {
       });
     });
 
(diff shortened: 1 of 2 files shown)


thanpolas:
Ok, i've rebased to master and fixed the tests so they use `commitMessage()`, they now pass.

Thank you for this!

webmakersteve:
It looks like you're instantiating a new consumer for each test, as it should, so I think it's fine. LGTM.

PR #143, opened 2017-03-21
Change stream implementation

webmakersteve:
TopicWritable was limited in its use. It took producers and a topic name
and allowed for a buffer -> kafka interface for an individual topic.

I like the simplicity, but a lot of the logic ought to be reused for
a much more common scenario: producing to variable
topics/keys/partitions.

Enter ProducerStream. Run it in objectMode and you can produce objects.
If it isn't in object mode, provide it a topic and get the base
functionality for free.

It also is now decoupled from a "parent" producer. Rather, it is created
with a factory method so the producer inside is owned by the stream.
This means we can react to stream events like `end` appropriately,
Connecting and disconnecting from kafka is as easy as spinning up a
stream and getting rid of it when you're done.


webmakersteve:
PR is open for some discussion. I will likely batch similar changes to the consumer stream interface before I release either of these two changes.

webmakersteve:
KafkaConsumerStream is now also included in this PR.

PR #165, opened 2017-04-29
Refactor configure step for compilation

webmakersteve:
Rather than relying on users to specify that they have SASL, we can use
librdkafka's built in checking to find out. If we parse out the header
file that is made we can export environment variables so node-gyp can
configure itself appropriately.

This commit should also improve development because configure is now run
at install time only. It means when developing you do not need to
recompile the entirety of librdkafka with each minor change

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,8 +1,8 @@
 {
   "variables": {
       # may be redefined in command line on configuration stage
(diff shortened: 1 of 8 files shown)


webmakersteve:
Do not merge yet. Just an idea and has largely been untested.

PR #185, opened 2017-06-01
Add support for offset commit callback

webmakersteve:
@edoardocomar


edoardocomar:
Great, thanks @webmakersteve and congrats on releasing v1.0.0 !
Looks like we'll need to add an e2e test. One day I will try that

Issue #189, opened 2017-06-07
Unable to produce message with an AVRO Key.


joelnet:
I have created a publisher and consumer using the kafka-rest service found here: http://docs.confluent.io/current/kafka-rest/docs/intro.html. This is working fine.

I am attempting to convert the producer to use node-rdkafka, but **I am finding the AVRO keys are being truncated**.

Example of a good message (using kafka-rest):
```
{"key":"AAAAAAEGbm93","value":"AAAAAALAy6+TCw==","partition":0,"offset":46500,"topic":null}
```

Example of a bad message, missing key (using node-rdkafka):
```
{"key":"","value":"AAAAAALy/LiTCw==","partition":0,"offset":46567,"topic":null}
```

A couple experiments:
```
// Buffer's
const encodedKey = encode(schema.key, keySchemaId, key);
const encodedValue = encode(schema.value, valueSchemaId, value);

// key is a Buffer and will be truncated
producer.produce('topic', null, encodedValue, encodedKey);
> {"key":"","value":"AAAAAALy/LiTCw==","partition":0,"offset":46567,"topic":null}

// key is a string and will be truncated
producer.produce('topic', null, encodedValue, encodedKey.toString('utf8'));
> {"key":"","value":"AAAAAALy/LiTCw==","partition":0,"offset":46567,"topic":null}

// still truncated
producer.produce('topic', null, encodedValue, '\x00\x00\x00\x00\x01\x06now');
> {"key":"","value":"AAAAAALy/LiTCw==","partition":0,"offset":46567,"topic":null}

// key is double base64 encoded (and invalid)
producer.produce('topic', null, encodedValue, encodedKey.toString('base64'));
> {"key":"QUFBQUFBRUdibTkz","value":"AAAAAALy/LiTCw==","partition":0,"offset":46567,"topic":null}

// key will be set to TEST
producer.produce('topic', null, encodedValue, 'TEST\x00\x00\x00\x00\x01\x06now');
> {"key":"TEST","value":"AAAAAALy/LiTCw==","partition":0,"offset":46567,"topic":null}
```

based on my last test, it seems that the key is being truncated at the first `\x00`, most-likely due to null-termination.

I am guessing that node-rdkafka is calling a C++ `produce` method that is calling `key->size()` instead of another overload that will allow you to set the size of the key. (I'm looking in ProducerImpl.cpp at RdKafka::ProducerImpl::produce).

[...]

PR #190, opened 2017-06-07
Allow keys to be submitted as buffers

webmakersteve:
Addresses #189


webmakersteve:
Currently this PR does not fix the issue of delivery reports. The only true fix to that will be to make all keys return as buffers, which will be a breaking change and need to wait until a future release.

webmakersteve:
Keys with null terminators must be specified using buffers - node.js casts its strings to c strings before they are entered into `std::string`, which can support extra null terminators.

PR #207, opened 2017-06-27
Updated gyp file for both librdkafka 0.9.5 and 0.11

edoardocomar:
Works with both librdkafka 0.9.5 and current master where new sasl files
need to be compiled.
Conditional for win32, but tested on macOS not Windows


webmakersteve:
Merged! Thanks for the PR.

PR #209, opened 2017-06-30
reenabled e2e test 'after assign, before consume, position should ...

edoardocomar:
reenabled e2e test 'after assign, before consume, position should return an array without offsets'


edoardocomar:
superseded by https://github.com/Blizzard/node-rdkafka/pull/210

PR #212, opened 2017-07-03
Fix probable typo in README

atamon:
`setInterval` takes milliseconds and not seconds I believe.

(Quite a silly change, but for someone not too familiar with JS this might cause a lot of confusion :) )

Solid work on a nodejs client btw. Good work!


webmakersteve:
Good catch.

PR #229, opened 2017-08-01
Make consumer example more verbose on config error

williamchong007:
... such as a missing or typo on `group.id`, which it just quit silently now


webmakersteve:
Sure thing, if it helps.

PR #231, opened 2017-08-02
Fixed README Document Inconsistency

d3r1v3d:
In the README example code, the `node-rdkafka` import examples alias to `kafka` (with a lower-case  k). Subsequent examples use a different convention -- pulling in libraries from `Kafka` (e.g. `new Kafka.Producer(...)`).

It's a minor inconsistency, but could lead to frustration with novice Node developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,8 +95,8 @@ This library includes two utility functions for detecting the status of your ins
 You can get the features supported by your compile of `librdkafka` by reading the variable "features" on the root of the `node-rdkafka` object.
 
 ```js
-const kafka = require('node-rdkafka');
-console.log(kafka.features);
+const Kafka = require('node-rdkafka');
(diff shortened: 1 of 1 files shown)


webmakersteve:
I'm wondering if we should fix it the other way - that is, make it so the documentation has it lowercased. Some linters complain when a Capitalized word is invoked as a non-class, and that may be equally frustrating ;) 

I think I'd prefer that. So, normalize it around `const kafka = require('node-rdkafka');`

Would you mind doing that instead? Then I'd be happy to merge it.

oliviertassinari:
> I think I'd prefer that. So, normalize it around const kafka = require('node-rdkafka');

💯  for using a lower case.

webmakersteve:
Closing because of inactivity.

Issue #45, opened 2016-10-20
node-rdkafka does not compile on Windows


webmakersteve:
Installing on windows does not work, and there is no way to make it work currently. 

This is a low priority feature, but keeping it here so it can be tracked properly.

serialLinker:
Hello, 

As you can see in this thread : https://github.com/edenhill/librdkafka/issues/982, librdkafka can now run on Windows platform (master at this time). Can you tell me when you think you can add this feature in node-rdkafka ?

I'm working for a major insurance company and node-rdkafka had been chosen for our node applications, and we prefer to use the same library on Windows and Linux.

Thx

mattness:
Windows support is not currently a priority for us. That said, we'd be happy to accept a pull request for it once a release of librdkafka has been made that includes edenhill/librdkafka#982.

vipreshjha:
@mattness  now that RC3 of librdkafka is released is there any possibility of providing a install for windows?

webmakersteve:
The reason it doesn't work on Windows is not because the compile of librdkafka didn't work on Windows: we just aren't able to prioritize its implementation.

As @mattness  said we are happy to accept a PR to implement the changes, but it isn't the librdkafka source that will be the work - it is adapting the gypfile to accommodate for all windows platform specific things, it is generating the header files required for the application to work (config.h) without assuming the presence of (ba)sh. It is testing the software on windows servers, and potentially integrating its testing into the CI pipeline.

I'd be happy to help anyone willing to take a crack at these problems so we could have windows support.

daves2552:
@webmakersteve Can you provide a very rough estimate of how many developer weeks you'd expect that effort to take?

vsimonian:
I'm currently working on Windows support for `node-rdkafka`. Right now it's building, but there are some linking issues to be resolved, and I don't have `WITH_SASL` working just yet.

However, it's only been a day so far that I've been working on it, so I expect to be making headway with some time.

RyanMelenaNoesis:
Just out of curiosity, is it even necessary to build librdkafka for Windows or would it suffice to simply use the [pre-built redistributable](https://www.nuget.org/packages/librdkafka.redist/)?

webmakersteve:
There is potential for that pre-built binary to work on Windows, but you would need to configure `node-gyp` to look for the library dependency there. It, too, would probably require changes to the code base to work, but may satisfy some of the complexity.

[...]

PR #236, opened 2017-08-14
Initial attempt at Windows support

RyanMelenaNoesis:
Uses [NuGet librdkafka - redistributable](https://www.nuget.org/packages/librdkafka.redist/) rather than compiling librdkafka for Windows.  Should address #45.

Also replaced `usleep` function with cross-platform `std::this_thread::sleep_for` alternative.

### DIFF
--- a/src/workers.cc
+++ b/src/workers.cc
@@ -378,7 +378,11 @@ void KafkaConsumerConsumeLoop::Execute(const ExecutionMessageBus& bus) {
       // when in consume loop mode
       // Randomise the wait time to avoid contention on different
       // slow topics
+	  #if defined(_WIN32)
       std::this_thread::sleep_for(std::chrono::milliseconds(static_cast<int>(rand() * 1000 / RAND_MAX)));
+      #else
+      std::this_thread::sleep_for(std::chrono::milliseconds(static_cast<int>(rand_r(&m_rand_seed) * 1000 / RAND_MAX)));
+      #endif
     } else if (
       b.err() == RdKafka::ERR__TIMED_OUT ||
       b.err() == RdKafka::ERR__TIMED_OUT_QUEUE) {
--- a/src/workers.cc
+++ b/src/workers.cc
@@ -378,14 +378,14 @@ void KafkaConsumerConsumeLoop::Execute(const ExecutionMessageBus& bus) {
       // when in consume loop mode
       // Randomise the wait time to avoid contention on different
       // slow topics
-      usleep(static_cast<int>(rand_r(&m_rand_seed) * 1000 * 1000 / RAND_MAX));
+      std::this_thread::sleep_for(std::chrono::milliseconds(static_cast<int>(rand() * 1000 / RAND_MAX)));
     } else if (
       b.err() == RdKafka::ERR__TIMED_OUT ||
       b.err() == RdKafka::ERR__TIMED_OUT_QUEUE) {
       // If it is timed out this could just mean there were no
       // new messages fetched quickly enough. This isn't really
       // an error that should kill us.
-      usleep(500*1000);
+      std::this_thread::sleep_for(std::chrono::milliseconds(500));
     } else if (b.err() == RdKafka::ERR_NO_ERROR) {
       bus.Send(b.data<RdKafka::Message*>());
     } else {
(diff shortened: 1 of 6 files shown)


src/workers.cc
```diff
@@ -378,14 +378,14 @@ void KafkaConsumerConsumeLoop::Execute(const ExecutionMessageBus& bus) {
       // when in consume loop mode
       // Randomise the wait time to avoid contention on different
       // slow topics
-      usleep(static_cast<int>(rand_r(&m_rand_seed) * 1000 * 1000 / RAND_MAX));
+      std::this_thread::sleep_for(std::chrono::milliseconds(static_cast<int>(rand() * 1000 / RAND_MAX)));
```
vsimonian: GNU `rand()` is not thread-safe, but `rand_r(*seedp)` is. Only MSVC's `rand()` is thread-safe. Therefore it seems that `rand()` should only be used in this context if it is being compiled by MSVC, not on all platforms.

RyanMelenaNoesis: Good catch @vsimonian.  C++ isn't my strong suit so thanks for the tip, I'll check in a change with an IFDEF momentarily.

mattness:
Thanks for the PR and your interest in the project, folks. We always love receiving contributions from the community!

Please remember that our first priority is ensure the stability and maintainability of the Linux build.  With that in mind, I have a couple objections to this PR as it stands right now:

1. `std:this_thread` is a C++11 construct.  `node-rdkafka` currently doesn't rely on C++11, and I'd prefer not to increase the compiler requirement if it can be avoided.
1. The addition of the `download` and `decompress` dependencies have no use in the library other than to support a platform-specific build. For applications that use node-rdkafka and don't run on Windows, these dependencies add additional attack surface area (as well as litter to the directory tree) with **zero** benefit to the application.
1. The `nuget-dep.js` and `node-gyp rebuild` steps in package.json are unnecessary for non-Windows installs, yet will still run and produce more litter. The reliance on nuget.org also introduces a requirement for internet connectivity that did not previously exist, and is commonly not readily available in enterprise environments or clean-room build environments.
1. I can't see anything that ensures the version of the binary being downloaded from nuget matches the revision of the source in the submodule. There's no way for a user of `node-rdkafka` to ensure that the downloaded binary doesn't contain malicious code, or for testers/maintainers to know that it is even built from the intended revision of librdkafka.

In general, I'm against downloading a pre-built binary (there's a reason node-gyp exists, after all). While it may be the path of least resistance to a "Works on my Machine" build, it's also a path of great burden for those who will be providing user support and maintaining the codebase.

I'd be more willing to accept a PR for Windows support where librdkafka was built from the submodule source at install time, just as the Linux build works now.

RyanMelenaNoesis:
With regards to your concerns:

1. I was working off the recommendation of @webmakersteve [here](https://github.com/Blizzard/node-rdkafka/issues/45#issuecomment-321307346) to avoid the need for a polyfill.  I would be open to surrounding the new imports and `std:this_thread` uses with IFDEFs or returning to the polyfill approach used by @vsimonian as well.
2. If you were open to it, I would be happy to convert the current nuget download/extract process into a bash script that could be executed in the npm scripts prepack step.  Alternatively, you could simply include the windows binaries as part of the repository.
3. Either of the alternatives I offered in point 2 would eliminate the need for internet connectivity for end-users.
4. There is nothing currently that controls the version of the binary being downloaded from nuget but it would be trivial to specify the nuget package version to download or use in the alternatives provided in point 2.  With regards to ensuring the downloaded binary doesn't contain malicious code, I think the wide-spread usage of NuGet, node-pre-gyp, pre-minified JS, CDNs, package managers, etc points to the concern being significantly overblown.

I'm happy to work further to address some of the concerns mentioned but requiring librdkafka to be built form source is frankly a bridge too far for me.  Given time constraints, if that is the bar that must be cleared I'd prefer to simply maintain my existing fork and point our package.json to it.

Personally, I'd have to think that non-ideal Windows support has to be better than no Windows support and I have a hard time seeing how the changes (even as implemented today) represent a 
> great burden for those who will be providing user support and maintaining the codebase

but obviously the final choice is not mine.

mattness:
Thanks so much for your effort thus far. We really do appreciate it. However, I need to insist that librdkafka is built from the submodule source.

I'll go ahead and close this PR for now, given that

> requiring librdkafka to be built from source is frankly a bridge too far

If time and inclination permits in the future, please feel free to reopen it with the needed changes.

PR #241, opened 2017-08-22
Update README.md

spezifant:
Removed wrong comma in parameter list.


webmakersteve:
Trailing commas are okay in JavaScript! Also, it looks like the real issue is that it is not including an opaque. I would prefer we add a parameter instead. If you'd like to do that I can take this PR.

Just have it so the next parameter is an arbitrary JavaScript object, if you'd like to make this change.

oliviertassinari:
The pull-request has been inactive for quite some time. I think that you can close it.

PR #244, opened 2017-08-27
Fix topic object breakages

cjlarose:
Commit a504c1e4611e680d33efbd51f115fde395967098 removed support for Topic objects, but they were still being created when writing to a producer stream.

Also, there was an integration test that explicitly tried to create Topic objects. This functionality is no longer supported, so I removed the failing test.


webmakersteve:
Just fixed this before I saw this PR :( Although I think it still has some gems I don't. I'll resolve conflicts and get this merged in :)

PR #254, opened 2017-09-25
Ensure ConsumeNum doesn't return early because of ERR_PARTITION_EOF

mimaison:
When hitting `ERR_PARTITION_EOF`, retry getting messages instead of breaking.
Also refactored the logic so it's similar to `KafkaConsumerConsumeLoop::Execute()`

Developed with @edoardocomar

### DIFF
--- a/src/workers.cc
+++ b/src/workers.cc
@@ -448,22 +448,27 @@ KafkaConsumerConsumeNum::KafkaConsumerConsumeNum(Nan::Callback *callback,
 KafkaConsumerConsumeNum::~KafkaConsumerConsumeNum() {}
 
 void KafkaConsumerConsumeNum::Execute() {
-  const int max = static_cast<int>(m_num_messages);
-  for (int i = 0; i < max; i++) {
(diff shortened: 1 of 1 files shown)


mimaison:
With this, when calling `consume(count, cb)`, we consistently get the expected number of messages. Otherwise, it would only get messages in chunks depending how librdkafka pre-fetched them.

webmakersteve:
Is there a reason you changed num messages from an int to a long? 

Lastly, I think with the new logic it makes more sense for it to be a switch case statement for readability.

edoardocomar:
Hi,
the change to long was to have this comparison compile
`m_messages.size() < max`

as for the if ... else rather than switch, we followed the `consumeLoop` pattern to make the two methods similar. We could switch them both to switch ?

webmakersteve:
I think consume loop was made to match this one, but now that it makes sense for this to be a switch we can make them both be switches. But i'll save that for a different PR - I can handle that :)

Thanks for the PR. merged!

PR #257, opened 2017-10-02
Fix Client.connectedTime method

philipyoo:
Small bug found while using the repo :) Please see below block for error message

```
/Users/philipyoo/Scality/wip/blizzard-kafka/node_modules/node-rdkafka/lib/client.js:243
  return Date.now() - this.metrics.connectionOpened();
                                   ^
TypeError: this.metrics.connectionOpened is not a function
```


webmakersteve:
Thanks for the PR!

PR #279, opened 2017-10-31
npm install now self contained

petermetz:
I wanted npm install to "just work" without me having to dance around it with scripts so I embedded the missing things in the package.json install script itself.

With this change the package works out of the box on node 8.x (8.0.0 and 8.8.1 were tested via the official docker image for NodeJS - note the alpine flavours don't work due to other issues with the compiler toolchain so you need the "proper" nodejs image not the alpine based)

I'm not sure if you wanted this fix in the first place or not, but figured I should give back so here it is.

Thanks for publishing the code, and much love to your work in general. I'm a longtime fan of Blizzard games.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "build": "node-gyp build",
     "test": "make test",
     "preinstall": "node util/configure",
-    "install": "node-gyp rebuild"
+    "install": "./configure; node-gyp rebuild"
   },
   "keywords": [
     "kafka",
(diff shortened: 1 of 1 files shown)


webmakersteve:
This shouldn't be necessary. Npm install will already install dependencies before it runs the gyp hooks, and configure is already run using the preinstall hook.

Are you having trouble installing it without making the changes you've made? Are you installing it from github or the npm registry?

webmakersteve:
I just removed the install hook override with b62ac1046ed7b39cee0b0e62742a5ce766d1a0ec. Can you try it again?

petermetz:
@webmakersteve Tested it just now, your change works equally fine for all my intents and purposes which makes my PR obsolete. Thank you.

Issue #294, opened 2017-11-14
ProducerStream error in produce function


michallevin:
Hi,

I am using the ProducerStream in objectMode. When writing more than 1 message to Kafka at the same time (without waiting for the previous write to finish), a `"topic" must be a string` error is thrown in the producer's produce function.

I noticed that in the first call to produce, the `topic` parameter exists and the `message` is of type Uint8Array. On subsequent calls, the `topic` parameter is `undefined` and the message is of type `Object` with the keys `partition`, `topic`, and `value` (all defined). But since the function's `topic` parameter is undefined, the error is thrown.

What is going on? 😨 
The code which configures the Producer stream and writes to Kafka is below for reference...

Thanks!  

```
const Kafka = require('node-rdkafka');
const EventEmitter = require('events');

let stream = null;

module.exports = function init(config, logger) {
  const kafkaEmitter = new EventEmitter();
  const hosts = config.kafka.hosts.reduce((host, string) => string.concat(',', host));

  stream = !stream ? Kafka.Producer.createWriteStream({
    'metadata.broker.list': hosts,
    'batch.num.messages': 50000,
    'message.send.max.retries': 5,
    'retry.backoff.ms': 5000,
    'message.max.bytes': 10 * 1024 * 1024,
    'request.required.acks': 1,
    'dr_msg_cb': true
  }, { 'request.required.acks': 1 }, {
    objectMode: true
  }) : stream;

  stream.producer.on('error', (err) => logger.error('Error writing to Kafka due to producer',
                                                        { error_message: err.message }));
  stream.on('error', (err) => logger.error('Error writing to Kafka due to stream', { error_message: err.message }));
  stream.producer.on('delivery-report', (error, message) => {
    return kafkaEmitter.emit(message.value, error)});

  return {
    write: write
  };

  function write(orgName, resource, value) {
    const message = {
      topic: `service-name.${orgName}.${resource}`,
      partition: 0,
      value: new Buffer(JSON.stringify(value))
    };

    const queuedSuccess = stream.write(message);

    if (!queuedSuccess) {
      logger.error('Kafka message not queued', { organization: orgName });
      return Promise.resolve();
    }

    return new Promise((resolve) => {
      kafkaEmitter.on(message.value, (error) => {
        if (error) {
          logger.error('Error in Kafka delivery report', { organization: orgName, message_value: value });
        }

        resolve();
      });
    });
  }
};

```

JaapRood:
[...]

PR #298, opened 2017-11-16
Fix draining of ProducerStream in object mode

JaapRood:
Experienced by myself and as reported by #294, a `ProducerStream` in `objectMode` being drained throws an error: `'topic' == undefined`.

I managed to add a test case that reproduces this issue. After adding another for a streams not in `objectMode` to make sure we don't regress, I implemented a simple fix, getting both tests to pass.

I'm new to the codebase and open to any suggestions for a different approach 😄 .

### DIFF
--- a/lib/producer-stream.js
+++ b/lib/producer-stream.js
@@ -201,6 +201,7 @@ function writev(producer, topic, chunks, cb) {
   // @todo maybe a produce batch method?
   var doneCount = 0;
   var err = null;
+  var chunk = null;
 
   function maybeDone(e) {
     if (e) {
@@ -213,7 +214,12 @@ function writev(producer, topic, chunks, cb) {
   }
 
   for (var i = 0; i < chunks.length; i++) {
-    producer.produce(topic, null, chunks[i], null);
+    chunk = chunks[i];
+    if (Buffer.isBuffer(chunk)) {
+      producer.produce(topic, null, chunk, null);
+    } else {
+      producer.produce(chunk.topic, chunk.partition, chunk.value, chunk.key, chunk.timestamp, chunk.opaque);
+    }
     maybeDone();
   }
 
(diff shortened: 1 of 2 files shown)


JaapRood:
Looking at why Travis is failing, it's only in Node v4 and doesn't seem to be at all related to the changes made here. Looks like `master` is failing them too?

michallevin:
@JaapRood This fix seems to work for us! We are hoping for a merged solution ASAP :)

bobrowadam:
@JaapRood I'm using your fix but am running into a queue is full issue. 
I'm trying to write messages in batches yet I fail to write in batches that are  > ~20 messages after `stream.write(message) === false`.
I believe this means that the queue is full although the default `queue.buffering.max.messages` is 100000.
Here is my config:
```
const stream = Kafka.Producer.createWriteStream({
    'metadata.broker.list': hosts,
    'batch.num.messages': 1000,
    'message.send.max.retries': 5,
    'retry.backoff.ms': 5000,
    'message.max.bytes': 10 * 1024 * 1024,
    'request.required.acks': 1,
    dr_msg_cb: true
  }, { 'request.required.acks': 1 }, {
    objectMode: true
  }
```
I was wondering if you saw anything similar.

JaapRood:
@bobrowadam can't say I'm familiar enough with the codebase to help you there. While having to do with queueing, it seems unrelated to fixing the draining issue we saw here. This was just a matter of `_writev` not being implemented for `objectMode: true`. Might be worth opening a separate issue for that.

JaapRood:
@webmakersteve is there anything I can do to help get this merged? The work around is pretty hacky, would be nice to get rid of it 😄 .

webmakersteve:
Sorry. Been a crazy few weeks.

webmakersteve:
Thanks for the PR!

JaapRood:
No worries @webmakersteve, happy to help!

JaapRood:
@bobrowadam I created pull request #398 that fixes a bug that caused a queuing error. It might be the one you've experienced.

PR #301, opened 2017-11-19
emit disconnected event on connect failure

evollu:
if getMetadata call fails, client disconnects itself silently. add a disconnect event so I can call connect again to reconnect.


webmakersteve:
Hmm.. I'm not sure how I feel about this.

I think a new event is justified here: `connection.failure`. Can you change it to that? `disconnected` is used only for user-initiated disconnections (because that's the only time you would ever be disconnected in the node client - though I admit it is a bit of a misnomer).

evollu:
done

PR #302, opened 2017-11-19
add auto reconnect example

evollu:
The client doesn't reconnect if connection to kafka fails in initial connection. I don't know if it is designed but I was expecting the client to retry.

adding an example code here


webmakersteve:
The client is not designed to automatically reconnect on an initial connection failure. It is up to you to check the error in the connection callback and act accordingly.

The example is fine, but I think it belongs either in a separate example file or in the README. I don't want people to think that is the only way to use the producer.

PR #342, opened 2018-01-23
Normalize API

oliviertassinari:
I want to restart my process in the event of a commit fail error.
In order to have access to the `err` object, I have to use the `offset_commit_cb` API.
I can't rely on the event emitted.

I'm not 100% it's the best approach to the problem. At least, **it's a breaking change**.

Also, I'm wondering if we shouldn't list `offset.commit` in the events table
> The following table lists important methods for this API.


webmakersteve:
It should be listed in the events table, but I wonder if there should be two separate events for this and we should check the error parameter before we emit the event.

I think I'd prefer that approach

PR #343, opened 2018-01-24
Additional return types

tonymrakovcic:
Added return types for:

connect
disconnect
getMetadata
queryWatermarkOffsets

For tsc to work properly if node-rdkafka is included in a project as a dependency

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,7 +3,7 @@
 export class Client extends NodeJS.EventEmitter {
     constructor(globalConf: any, SubClientType: any, topicConf: any);
 
-    connect(metadataOptions: any, cb?: (err: any, data: any) => any);
+    connect(metadataOptions: any, cb?: (err: any, data: any) => any): any;
 
     getClient(): any;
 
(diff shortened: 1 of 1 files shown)


webmakersteve:
I believe those methods return the Client object, not an any. They can be chained that way.

tonymrakovcic:
Thank you, correct me if I'm wrong but only connect and disconnect return Client
disconnect  - Client
https://github.com/Blizzard/node-rdkafka/blob/e03bf79e1f5ccfbea900e8c8eae06aae03eeac63/lib/client.js#L279

connect - Client
https://github.com/Blizzard/node-rdkafka/blob/e03bf79e1f5ccfbea900e8c8eae06aae03eeac63/lib/client.js#L128

These two return:
getMetadata - void, or return cb(new Error('Client is disconnected'));
https://github.com/Blizzard/node-rdkafka/blob/e03bf79e1f5ccfbea900e8c8eae06aae03eeac63/lib/client.js#L329

queryWatermarkOffsets - void | return cb(new Error('Client is disconnected')) | 
https://github.com/Blizzard/node-rdkafka/blob/e03bf79e1f5ccfbea900e8c8eae06aae03eeac63/lib/client.js#L366

Which could be "any"

I have edited the types in that fashion now

PR #353, opened 2018-02-13
remove dead comment

oliviertassinari:
The comment header seems to be duplicated.


webmakersteve:
That function overload is still true. There is an overload for the consume function, where, when you just pass it a callback, it will keep the loop pumping constantly until disconnection.

oliviertassinari:
Oh my bad. So the double comment is just here to express it's a variadic function.

PR #368, opened 2018-02-28
Update README.md

KoryNunn:


webmakersteve:
Can you put this in the installation steps rather than usage? And it also probably isn't good to put the actual package name because it differs on other linux distributions.

PR #378, opened 2018-03-15
Add TypeScript definitions for 'close' methods

dchesterton:


webmakersteve:
Is there a way we can specify the definition for the callback? And the return value for close should be undefined.

dchesterton:
It looks like this has now been fixed in master so I'll close the PR.

PR #379, opened 2018-03-18
Add missing producer method

spasam:
@webmakersteve I was trying to use `queryWatermarkOffsets` on `producer` and it was erroring out.


alexayalabonilla:
I have the same issue, consumer side is working fine. 

Consumer throws the next error:

```
TypeError: this._client.queryWatermarkOffsets is not a function
client.js:386
    at Producer.Client.queryWatermarkOffsets (/node_modules/node-rdkafka/lib/client.js:386:16)
    at producer.connect (kafkaProducer.js:60:16)
    at next (/node_modules/node-rdkafka/lib/client.js:134:7)
    at /node_modules/node-rdkafka/lib/client.js:213:7
    at /node_modules/node-rdkafka/lib/client.js:348:7
```

webmakersteve:
Generally used in the consumer, which is why I put it there - but no harm in including it in the producer too.

spasam:
@webmakersteve Thanks. I was confused because of documentation and was trying to use it on producer.

PR #381, opened 2018-03-20
Accept `connectOptions` with ProducerStream constructor

JaapRood:
Mirroring the behaviour of the `ConsumerStream`, this adds the ability to forward connection options when creating a ProducerStream. Very useful when trying to control the client's `timeout` option, amongst other things.

I added a simple test verifying the option is being passed down from the constructor. I'm happy to make any other changes if required.


webmakersteve:
Nope. This is perfect :)

Issue #344, opened 2018-01-24
TypeError: Cannot read property 'isConnected' of null


oliviertassinari:
I ended up in the following edge case:
1. I close the producer stream
2. I try to write in the producer stream
3. An error is throw: `TypeError: Cannot read property 'isConnected' of null` from this line: https://github.com/Blizzard/node-rdkafka/blob/22c713b2d9dfbb2b0ef9d42bf2b890c4ed9e4133/lib/producer-stream.js#L121

This is the expected behavior given the producer is cleared:
https://github.com/Blizzard/node-rdkafka/blob/22c713b2d9dfbb2b0ef9d42bf2b890c4ed9e4133/lib/producer-stream.js#L268
It's an edge case, let me know if you want to support it. Eventually it could be simply about raising an explicit error. Hopefully this issue description will help people facing it.

oliviertassinari:
Oh, it's an already existing pattern:
https://github.com/Blizzard/node-rdkafka/blob/3a2d5835602806f364108cc2fae338e6615a98fe/lib/producer.js#L166

webmakersteve:
It shouldn't throw a null pointer, so that needs to be adjusted if it is happening. I would prefer it throwing a more specific exception. So it's definitely a bug!

webmakersteve:
Fixed with https://github.com/Blizzard/node-rdkafka/commit/a864340ef345b721414ec9be6245b3064ddae738

No need for the producer to be set to null. I did it previously because the stream has undefined behavior once the producer internally has been closed, but I also want to provide the flexibility of accessing the internal producer after the stream has been closed (e.g. to get metrics, etc..)

So, I took the set to null out.

oliviertassinari:
:+1:

PR #384, opened 2018-03-23
Prevent KafkaConsumerStream from losing it's KafkaConsumer reference

JaapRood:
As with #344, I ran into an issue with the `KafkaConsumer` reference being lost from the `KafkaConsumerStream` after closing the stream while it's reading at a high rate. 

Through the asynchronous nature of a `Stream`, it's possible there's some left to read even if the underlying consumer has been disconnected. A call in the `_read` methods to check whether the consumer is connected throws an error, as the consumer was set to null.

I don't believe that removing the reference does anything in the first place? Happy to be made of aware of anything, so I can help mitigate it some other way.

### DIFF
--- a/lib/kafka-consumer-stream.js
+++ b/lib/kafka-consumer-stream.js
@@ -358,7 +358,6 @@ KafkaConsumerStream.prototype.close = function(cb) {
   if (self.consumer._isConnected) {
     self.consumer.unsubscribe();
     self.consumer.disconnect(function() {
-      self.consumer = null;
       close();
     });
   }


webmakersteve:
The only thing setting it to null provides is allowing all the memory to be garbage collected when the stream is done processing. But it may be more important to keep a handle open to that object to allow post-close interaction.

JaapRood:
@webmakersteve thanks for the merge! Any chance we could get the last couple of PRs released and published to npm? We're still looking at `2.3.1` as the latest.

webmakersteve:
2.3.2 has been published - happened 10 days ago I think?

PR #386, opened 2018-03-26
Fix extended classes for ProducerStream and ConsumerStream

jacobparra:
According to https://github.com/Blizzard/node-rdkafka/blob/2b98bc36ac77088b04bb0bba76056b4ea0f81013/lib/kafka-consumer-stream.js#L14 and https://github.com/Blizzard/node-rdkafka/blob/2b98bc36ac77088b04bb0bba76056b4ea0f81013/lib/producer-stream.js#L14 the extended class is from the `stream` lib. This allows to call the `destroy()` function in both objects when using TypeScript.


webmakersteve:
I don't use typescript much, but is NodeJS.WriteStream different than what you're doing there?

jacobparra:
I've been making some experiments with this lib and typescript. When trying to transpile to JS the linter throws the error `Property 'destroy' does not exist on type 'ConsumerStream'.`. 

Looking at the definition for `NodeJS.ReadStream` with the editor, it goes to this line: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/38bda769b62d2583bf14ad37db854460b833bbfa/types/node/v8/index.d.ts#L539 which extends from `Socket` on the `NodeJS` namespace. The same happens for `WriteStream`.

That's why I was looking for the exact definitions for `Readable` and `Writeable` classes on the `stream` module. This classes I think corresponds to the ones required in this lib.

webmakersteve:
Makes sense. Thanks for the explanation.

Issue #401, opened 2018-04-13
KafkaConsumerStream consumption rate causing time-outs and overall instability on clusters with throughput quota


JaapRood:
We're running a service that consumes messages from a cluster with [a throughput quota](https://cwiki.apache.org/confluence/display/KAFKA/KIP-13+-+Quotas). At moments of high throughput we're getting reports of throttling through `event.throttle`, with the amount the broker throttled a Fetch request going up steeply and reaching time-outs.

As a result, everything becomes unstable. The most common thing we see is a `ConsumerGroup` heartbeat timing out, causing a rebalance event. As we restore processing state from Kafka this only puts more stress on the cluster.

I've tried implementing a `KafkaConsumerStream` based solution, listening for `event.throttle` from the consumer and wrapping `_read_message` and `_read_buffer` to delay consumption for a while. However, it seems that because `KafkaConsumer.consume()` uses it's own timeout, we never wait for the throttled response from the broker to come back, basically ignoring the back-pressure the Broker is trying to give. In our testing this means we've generally already made ~30 requests for the first throttle event from request~10 to come in, with it already to late to react to.

## Expected behaviour

For the consumption of messages to respect throughput quotas set, slow down consumption and prevent time-outs causing more instability.

## Possible fix

I'm not sure how `librdkafka` deals with quotas itself, if at all. My limited knowledge of C and just the reference docs don't paint me a picture. That makes it hard to say at what level of the stack a back-off would be implemented, as it already might be at some point.

_If_ the `KafkaConsumerStream` is the place to implement this (which I can imagine as it already is responsible for the back-pressure mechanism), it would need feedback from calling `consumer.consume()` about the consume call having timed-out. It could then respond by backing-off future consumption until either a subsequent request doesn't time out or throttling information is received. Currently, from what I can gather, [the required time-out errors are caught by the worker executing the consumption call and not forwarded to Node](https://github.com/Blizzard/node-rdkafka/blob/master/src/workers.cc#L523-L550).

I'd be more than happy to give the `KafkaConsumerStream` level of rate limiting a shot if that's the way to go. I just can't really tell if it is 😅.

webmakersteve:
I think having the stream implement `event.throttle` is a great idea, and is the perfect place to put that logic. I reviewed your PR initially and left some other comments there.

PR #404, opened 2018-04-17
Back off ConsumerStream's consumption exponentially upon receiving time outs

JaapRood:
This is a proposal for a fix to #401. It forwards consumption time-out errors to the Node.js `KafkaConsumer`, which lets `KafkaConsumerStream` handle it and back off consumption. A couple of notes:

- Before `ERR__TIMED_OUT` and `ERR__TIMED_OUT_QUEUE` were caught and never made it to Node, meaning `KafkaConsumer.consume` could now return an error for code where it didn't before. This could be interpreted as a breaking change.
- To implement the exponential back-off I've added [`retry`](https://www.npmjs.com/package/retry) as a dependency. That was mostly to focus on the effect of the back-off, rather than implementing an effective back-off mechanism (which, [as this article describes well](http://dthain.blogspot.com.au/2009/02/exponential-backoff-in-distributed.html), isn't totally trivial). 
- While it's solved the issue for which I created #401, this back-off is rather "blind". It doesn't use the [`throttleTime` ](https://cwiki.apache.org/confluence/display/KAFKA/KIP-13+-+Quotas#KIP-13-Quotas-PublicInterfaces) field as returned by the Broker. If anyone has some good concrete ideas about how to incorporate that, I'm all ears, because it would be a lot more effective.

### DIFF
--- a/lib/kafka-consumer-stream.js
+++ b/lib/kafka-consumer-stream.js
@@ -13,6 +13,8 @@ module.exports = KafkaConsumerStream;
 
 var Readable = require('stream').Readable;
 var util = require('util');
+var ErrorCode = require('./error').codes;
+var Retry = require('retry');
 
 util.inherits(KafkaConsumerStream, Readable);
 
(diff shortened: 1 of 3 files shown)


JaapRood:
@webmakersteve I'd really appreciate it if you could shed your light on this. I'm not familiar with the codebase back to front, so it's very possible this approach is not a great one. I'll hold off on making more progress on it in the meantime.

JaapRood:
An alternate to this back-off mechanism could be to raise the time-out time of the `KafkaConsumer.consume` over time and monitoring the reported throttle time from the Broker. The advantage of that would be that the consumer (especially a single one) would get closer to the actual throughput limit. That's at the same time also the disadvantage: it would still pressure the Broker and get you close to the cascade of timeouts described in #401.

JaapRood:
Looking into why the tests aren't passing, I manage to uncover that time-out errors were once forwarded to the Node.js side of things. It seems that #103 is where the switch to handling / swallowing the errors is made, as a bigger change of cleaning up the consumption API. @webmakersteve since you were there and the author of that PR, I'm hoping you can tell us if we're missing anything in particular.

webmakersteve:
You are correct - timeout errors did not cause a corresponding "node" error upstream. The reason for this decision is that a timeout reading more often than not is just a lack of messages in the log. The consume call will wait for messages to be produced on the partitions it is assigned, and does not exit immediately if no messages are in the log.

This did not seem like an error case for me in node land. In the C/C++ land, it is technically an error code but only because you need to be notified that the call did not result in a message to be read. Sometimes it is indeed an error and deserves special handling (perhaps in the case of retries), but a lot of times it doesn't.

I know in some of our code we have consumers that get very few messages a day, and I don't want timeouts to cause finite numbers of retries, or even backoffs on the consumer.

I think for a lot of use cases this makes sense, but some special thought will need to go in to make this configurable, and a lot of changes across the code base will need to be done to accommodate for the now error case that is a consume timeout, where before it was just treated as "there is no data to read"

It may be better to hold off on this until an error case pass can be done. Honestly I'd prefer that all errors bubbled up in some capacity and that the JS layer could handle interpreting them if needed, rather than leaving that logic in C/C++. This will allow more flexibile approaches and keep the C++ layer as a 1:1 bindings layer for the library. But then something like this could be built on top.

JaapRood:
@webmakersteve great! I'll try and fit it in this week, as I'll be working on a different project next week, not coming back to our Kafka / stream processing stuff until the week after (small team, limited resources 😇).

JaapRood:
@webmakersteve I ran out of time this week, so it'll be another 1 or 2 weeks before the next opportunity for me to pick this up. Feel free to tackle it yourself if that's too far away and has you waiting 😬.

JaapRood:
I'm going to close this, as merging it would have no effect. Since I first introduced this, I've learned how `librdkafka` actually fetches data and deals with broker throttling for us. There were other issues that caused the time-outs we've seen, mostly triggered through running a high amount of consumers in addition to performing other consumer operations like `seek`, `pause` and `etc`. #554 solves the root of these issues a lot better.

PR #407, opened 2018-04-20
Add typescript definition overload for both consumer modes.

pthm:
Previously only supported the non-flowing option and would complain
that arguments were missing.

This allows TS to compile when using flowing mode


wong2:
👏 waiting for this

PR #420, opened 2018-05-21
types: Typing Client & KafkaConsumer

alexander-alvarez:
### Changes
- Replacing `any` with `this` in cases where `this` is actually being returned
- added typedef for `commit()` with no arguments
- `subscribe(topics)` should be `topics: string[]` not `topics: any`
- connect has a no-argument invocation, so defining `connect(metadataOptions? any)`
- `consume(number)` should be a `number: number` not `number: any`
- `T extends Repository` stuff -- allows us to declare specific types so users have access to them when they are defining callbacks, instead of simple "any"

so for example when defining
```ts
consumer.on('data', (message) => {
  // TS will know `val` is a Buffer & `message` is of type ConsumerStreamMessage
  let val = message.value;
});
```

## Help
- I left a lot of the callbacks in the repository as `any` since its better that they be flexible instead of wrong, but would love help in defining them (either in this PR or subsequent)

See https://github.com/Blizzard/node-rdkafka/compare/master...alexander-alvarez:more-typings?expand=1#diff-b52768974e6bc0faccb7d4b75b162c99R193

- Similarly I wrapped the return types of the methods that returned `this._errorWrap(this.client[method], ...)` with `ErrorWrap<T>` so that we could better highlight the intent (although `boolean | T` is not necessary when the `T` is `any`). I need help refining those return types in subsequent PRs also.


alexander-alvarez:
@webmakersteve could I get a review here?

PR #421, opened 2018-05-22
Fix memory leak in ToV8Object

DarrenTom:
Allocate memory using Nan::Encode to ensure handle garbage collection frees it as expected.


DarrenTom:
Hold off review of this change.
I'm looking into the test failure.

DarrenTom:
Ok.  I just needed to encode to a Buffer.  You may proceed with the review.

PR #440, opened 2018-07-03
Don't require KafkaConsumer to be subscribed to pause and resume the consumption of toppars

JaapRood:
I ran into this limitation while trying to use `consumer.pause` with a `KafkaConsumer` that isn't subscribed to a Consumer Group, but assigned toppars manually. In neither rdkafka's or the original Java implementation and documentation can I find anything to suggest that this should prevent the use of these control flow methods, but I'm happy to hear what I'm missing :)

### DIFF
--- a/src/kafka-consumer.cc
+++ b/src/kafka-consumer.cc
@@ -358,7 +358,7 @@ Baton KafkaConsumer::Unsubscribe() {
 }
 
 Baton KafkaConsumer::Pause(std::vector<RdKafka::TopicPartition*> & toppars) {
-  if (IsConnected() && IsSubscribed()) {
+  if (IsConnected()) {
     RdKafka::KafkaConsumer* consumer =
       dynamic_cast<RdKafka::KafkaConsumer*>(m_client);
     RdKafka::ErrorCode err = consumer->pause(toppars);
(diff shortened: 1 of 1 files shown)


webmakersteve:
Without a subscription set, pause and resume probably won't do anything, but if it doesn't hurt anything it doesn't really matter to me either way.

Have you tested running this when not subscribed? Is it possible to write an end-to-end test that tries that behavior and ensures it does not crash?

JaapRood:
> Without a subscription set,  pause and resume probably won't do anything [...]

You can assign toppars manually to a `KafkaConsumer`, rather subscribing to the group, which is where I ran into this. In this case, it's for the same reason as this example of [the official KafkaConsumer docs](https://kafka.apache.org/0102/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#manual-partition-assignment):

> In the previous examples, we subscribed to the topics we were interested in and let Kafka dynamically assign a fair share of the partitions for those topics based on the active consumers in the group. However, in some cases you may need finer control over the specific partitions that are assigned. For example:
> - If the process is maintaining some kind of local state associated with that partition (like a local on-disk key-value store), then it should only get records for the partition it is maintaining on disk.

Even before assignments are set through `consumer.assign`, either manually or through a subscription, you might want to pause any message consumption for it. In my case, as part of receiving the assignment historical data has to be restored before processing can resume.

When I wrote this fix I ran past a librdkafka issue with confirmation from it's author that it's a supported use case, but it's proving tricky to find back now 😇. I'm happy to write an end-to-end test to confirm this empirically, once I get the chance!

Issue #405, opened 2018-04-17
Error: librdkafka++.so.1: cannot open shared object file


JaapRood:
Reposting of #394, as it was closed by the issue author without being fixed.

When trying to use run on Heroku (and I'm guessing other linuxed-powered server environments), the following error is triggered trying to require `node-rdkafka@2.3.2`:

```
Error: librdkafka++.so.1: cannot open shared object file: No such file or directory
    at Object.Module._extensions..node (module.js:681:18)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at bindings (/app/node_modules/bindings/bindings.js:81:44)
    at Object.<anonymous> (/app/node_modules/node-rdkafka/librdkafka.js:10:32)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
```

The author of #394 suggested adding an Aptfile to your project, however, this is non-customary for npm packages. That said, I can't exactly suggest a solution myself, being really unfamiliar with how Node.js binds to C/C++, the build systems work, etc.

alexander-alvarez:
I had this issue when building and moving assets from one docker image to another...
which we initially needed to do because we were using `node-slim` which doesn't have all of the libraries to build `librdkafka`. Building it on the regular node image, and not copying it over to a different container from where it was built fixed the issue for me.

JaapRood:
Thanks for your input @alexander-alvarez, that puts some more weight behind this being a portability issue, most likely through the use of an absolute path rather than a relative one.

webmakersteve:
Does Heroku not move the entire node_modules directory? I see people have been having problems binding dynamically to those libraries since the build change. I don't use Heroku so things aren't tested there, but if anyone has any insight into this issue would happily take a contribution to fix this issue for Heroku users

battlecow:
@webmakersteve I run my app in docker and copy the entire node_modules directory over to the final image, I have had to go about deleting this lib directory and re-running npm install on image build.

JaapRood:
@webmakersteve just like @battlecow does with his app in Docker, Heroku builds from project root and then copies the project root and all of it's contents to different locations. More specifically, it creates a compressed slug from the result of the build, so that it can easily provision new (but identical) servers with new instances, having to only build the project once.

[...]

PR #442, opened 2018-07-09
feat: Using relative paths in builds

alexander-alvarez:
Here's me taking a stab at getting relative path builds working in the linux environment in response to portability issues brought up in https://github.com/Blizzard/node-rdkafka/issues/405

This was mostly implemented though guess and check, so I am more than open to feedback. I got rid of every instance of `<(module_root_dir)/` except for the one in the `include` (which I couldn't get to pass without it).

I'm not sure which paths could be absolute, which need to be relative, but I wanted to get the ball rolling so I converted everything I could.

The mac build we'll have to iterate on with travis since I only have access to a linux environment

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -122,10 +122,10 @@
                   'xcode_settings': {
                     'MACOSX_DEPLOYMENT_TARGET': '10.11',
                     'GCC_ENABLE_CPP_RTTI': 'YES',
-                    'LDFLAGS': [
+                    'OTHER_LDFLAGS': [
                       '-L/usr/local/opt/openssl/lib'
                     ],
-                    'CPLUSPLUSFLAGS': [
+                    'OTHER_CPLUSPLUSFLAGS': [
                       '-I/usr/local/opt/openssl/include',
                       '-std=c++11'
                     ],
--- a/binding.gyp
+++ b/binding.gyp
@@ -37,7 +37,7 @@
                   'librdkafkacpp.lib'
                 ],
                 'AdditionalLibraryDirectories': [
-                  '<(module_root_dir)/deps/librdkafka/win32/outdir/v120/x64/Release/'
+                  'deps/librdkafka/win32/outdir/v120/x64/Release/'
                 ]
               },
               'VCCLCompilerTool': {
@@ -45,7 +45,7 @@
                   '/GR'
                 ],
                 'AdditionalUsingDirectories': [
-                  '<(module_root_dir)/deps/librdkafka/win32/outdir/v120/x64/Release/'
+                  'deps/librdkafka/win32/outdir/v120/x64/Release/'
                 ],
                 'AdditionalIncludeDirectories': [
                   'deps/librdkafka/src-cpp'
@@ -55,7 +55,7 @@
             'msvs_version': '2013',
             'msbuild_toolset': 'v120',
             "dependencies": [
-              "<(module_root_dir)/deps/librdkafka.gyp:librdkafka"
+              "deps/librdkafka.gyp:librdkafka"
             ],
             'include_dirs': [
               'deps/librdkafka/src-cpp'
@@ -66,7 +66,7 @@
               [ "<(BUILD_LIBRDKAFKA)==1",
                 {
                   "dependencies": [
-                    "<(module_root_dir)/deps/librdkafka.gyp:librdkafka"
+                    "deps/librdkafka.gyp:librdkafka"
                   ],
                   "include_dirs": [
                     "deps/librdkafka/src-cpp"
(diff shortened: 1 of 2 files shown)


alexander-alvarez:
@battlecow I don't think those environment variables work

From the logs in the build:
https://travis-ci.org/Blizzard/node-rdkafka/jobs/402243743
```
 LDFLAGS                  -L/Users/travis/build/Blizzard/node-rdkafka/build/deps
 CFLAGS                   -I/usr/local/Cellar/openssl/1.0.2h_1/include
```
instead of the overridden ones you provided

battlecow:
@alexander-alvarez I actually removed those completely just now from the file and I can build fine apparently.  I had only brought them in since you had them specified in your other branch.

alexander-alvarez:
Hmm... the `OTHER_LDFLAGS` don't seem to show up in the `LDFLAGS` section though but I've seen it in the wild with: https://github.com/sass/node-sass/blob/master/binding.gyp


Maybe we consider doing something like: 
https://github.com/KhurshidAlam/https-github.com-joyent-node/blob/master/node.gyp#L211
It seems they add openssl to their `deps` folder

battlecow:
Yea I had been having the same issues with requiring those overrides to make the build work, but I just rolled back to LTS via brew so that is probably what fixed the linking issues for me at least.  I'll keep seeing if I can re break my setup to test those flags.

alexander-alvarez:
>  I just rolled back to LTS

rolled what back to LTS? openssl? how did you do this? I have access to coworkers with "broken setups" who I can guinea pig with any steps you have to see if that fixes it for them

battlecow:
sorry, LTS node 8.11.3 from 10.6.0.  I just uninstalled openssl and reinstalled it to no effect as well as moving back to 10.6.0.  Here's my openssl info from brew:

``` ➜  mac-node-kafka git:(de12747) brew info openssl
openssl: stable 1.0.2o (bottled) [keg-only]
SSL/TLS cryptography library
https://openssl.org/
/usr/local/Cellar/openssl/1.0.2o_2 (1,792 files, 12.3MB)
  Poured from bottle on 2018-07-10 at 10:44:26
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/openssl.rb
==> Dependencies
Build: makedepend ✘
==> Options
--without-test
	Skip build-time tests (not recommended)
==> Caveats
A CA file has been bootstrapped using certificates from the SystemRoots
keychain. To add additional certificates (e.g. the certificates added in
the System keychain), place .pem files in
  /usr/local/etc/openssl/certs

and run
  /usr/local/opt/openssl/bin/c_rehash

This formula is keg-only, which means it was not symlinked into /usr/local,
because Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries.

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/openssl/bin:$PATH"' >> ~/.zshrc

For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/openssl/lib
    CPPFLAGS: -I/usr/local/opt/openssl/include
For pkg-config to find this software you may need to set:
    PKG_CONFIG_PATH: /usr/local/opt/openssl/lib/pkgconfig
```

alexander-alvarez:
Oh got it. We're seeing issues with node `v8.9.4`, and now that you mention it the clean install I did yesterday (that I couldn't break) would have picked up the LTS, so maybe it's related. Investigating.

battlecow:
So I can get it to break when doing a normal npm install from release, but not when I clone the repo and run it from there locally so I am not clear on how I can replicate it and still be able to modify the files so it breaks in a reliable way to fix it.

alexander-alvarez:
yeah that wasn't it... :(

battlecow:
How did you setup the repo environment on the Mac? I cloned your fork, cloned eden kafka into deps/librdkafka, checked out v0.11.4, then ran the npm install from root.

alexander-alvarez:
I followed https://github.com/Blizzard/node-rdkafka#tests for the submodules

alexander-alvarez:
@webmakersteve any insights?

battlecow:
To my inexperienced self it looks like nothing I am changing in the gyp file is being passed into the lib kafka build which is causing the issues.  I am quite unfamiliar with all of this so trying all of these combinations locally doesn't appear to be producing any results unfortunately.

webmakersteve:
LGTM. Gonna give it a shot :) 

Thanks for the contribution! Sorry I'm so late on responding.

vaibhav93:
Thanks for fixing this. Can this be released in 2.3.5 anytime soon?

shubhi1407:
I am facing the same issue, Can fix be released by Monday?

webmakersteve:
Releasing this today. Was trying to see if I could get some more in but time has not been on my side.

Issue #460, opened 2018-07-31
Do not modify config objects


rmdm:
I have encountered the following issue: when several producers initialized with the same config object (by reference) only the first one receives delivery report. But when I create new config object out of the original one (e.g. with Object.assign) for each producer, everything seems ok. Here is the code to reproduce the issue:

```javascript
const Kafka = require('node-rdkafka')

const config = {
    'metadata.broker.list': 'localhost:9092',
    'dr_cb': true,
}

async function runExample () {

    const producer1 = await getProducer(config)
    const producer2 = await getProducer(config)
    console.log('producers created')

    const delivery1 = awaitDelivery(producer1)
    const delivery2 = awaitDelivery(producer2)

    producer1.produce('topic1', null, Buffer.from('m1'))
    producer2.produce('topic2', null, Buffer.from('m2'))
    console.log('messages produced')

    await delivery1
    console.log('delivery1')

    await delivery2     // <-- hangs here
    console.log('delivery2')

    await disconnect(producer1)
    await disconnect(producer2)
}

function getProducer (config) {

    const producer = new Kafka.Producer(config)
    // const producer = new Kafka.Producer(Object.assign({}, config))   <-- if changed to this, all works

    producer.setPollInterval(100)

    return new Promise(function (resolve, reject) {
        producer.connect({}, function (err) {
            if (err) { return reject(err) }
            resolve(producer)
        })
    })
}

function awaitDelivery (producer) {
    return new Promise(function (resolve, reject) {
        producer.on('delivery-report', function (err) {
            if (err) { return reject(err) }
            resolve()
        })
    })
}

function disconnect (producer) {
    return new Promise(function (resolve) {
        producer.disconnect(5000, function () {
            resolve()
        })
    })
}

runExample()
.then(function () {
    console.log('done')
}, function (err) {
    console.error(err)
    process.exit(1)
})
```

**node-rdkafka** is of **2.3.4** version

The reason for this is that the library modifies passed config objects, by deleting or adding properties in both `KafkaConsumer` and `Producer` constructors. In the example, in particular, it drops `dr_cb` property. I think that creating a copy of the passed config object, instead of messing with the original one would be much more friendly behaviour.

PR #464, opened 2018-08-04
Do not modify passed config objects. Make a shallow copy instead.

rmdm:
Fixes: #460


webmakersteve:
Thanks! Always good to not modify the object :)

Issue #467, opened 2018-08-08
KafkaConsumer~consume and KafkaConsumerConsumeNum::KafkaConsumerConsumeNum have different interpretations of the timeout/partition end


ankon:
In our environment we have very few messages per second coming into our topics, and we want to process these as quickly as we can. 

We noticed a considerable latency of 1s minimally (never less than that), and up to a few seconds when using node-rdkafka. Testing with standard kafka and rdkafka performance tools didn't exhibit any issues and showed us that the cluster and the client are capable of easily handling the amount of messages in our desired latency. Further instrumentation of our own producer code also showed that the latencies of producing messages (time between us handing the message off and receiving the delivery report for that message) where acceptable.

We then tested with a combination of the `rdkafka_performance` tool as producer of messages that would be acceptable to our system, and measured the latency there: We still observed the 1s latency. In one instance I changed the rate of producing messages to 1/s, and actually got to a state where our consumer would not process any message for multiple seconds, and kafka would report the lag on the consumer group to increase continuously.

Enabling suitable debug logs of rdkafka (aka "all" :D) showed that while rdkafka reports it is fetching the new messages, it never seemed to hand them over to us.

Reading through the code I foundt that the problem is in KafkaConsumerConsumeNum::KafkaConsumerConsumeNum: When it reaches the end of the partition it simply retries rather than aborting, and for each retry it applies the timeout again. In the worst case this leads to the situation that it reads one message, hits the end-of-partition, retries reading with a timeout of 1s, and just before the timeout ends receives another message. As long as the number of messages doesn't exceed the requested maximum it will then go on, ending in a potentially "infinite" loop if the producer just sends out a slow stream of messages.

In our case we call consume with a size of 1000, and a timeout of 1000, the idea being to get "a lot of messages" when there are plenty, and avoiding busy loops when there are no messages.

KafkaConsumer~consume(sz, cb) states:
> This will keep going until it gets ERR__PARTITION_EOF or ERR__TIMED_OUT
> so the array may not be the same size you ask for. The size is advisory,
> but we will not exceed it.

This sounded reasonable for me, I assumed that it would get us whatever it has in big batches (sz=1000), but would quickly stop when the partition has reached the end.

I've now changed the timeout to 10ms, which gets us nice latencies, but I think something needs to be done in KafkaConsumerConsumeNum:
1. Fix the handling of the partition end to actually abort (matching the description of the caller), and/or
[...]

PR #469, opened 2018-08-09
 ConsumeNum returns messages on partition EOF

Tapppi:
Modify ConsumeNum to return messages on partition EOF, to match documented behaviour of `consume(num, cb)` as described in https://github.com/Blizzard/node-rdkafka/issues/467. Regardless of confusion, this would improve latency when fully draining partitions which is the default state in non-bursty applications.

I think EOF behaviour should at least be changeable with an option, like default consume timeout is. Would also be useful to expose these in stream implementation to allow fine-tuning with it. A batch stream can be tuned quite well and the overhead isn't that big.

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -14,6 +14,7 @@ var Kafka = require('../');
 var kafkaBrokerList = process.env.KAFKA_HOST || 'localhost:9092';
 var eventListener = require('./listener');
 var topic = 'test';
+var topic2 = 'test2';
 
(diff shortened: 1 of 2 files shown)


Tapppi:
@webmakersteve Any thoughts about this?

webmakersteve:
Took a look at it. I think this is a good half measure to get it where I'd like to get it in the future (where, it would quit out and return the error code). 

So I'll merge it to help assist with the use cases you're describing.

Issue #413, opened 2018-05-07
node-rdkafka build fails in Windows


indrajithpatelrsap:
Hi Colleagues,

I have been trying to install node-rdkafka library from past 3 days, but the Build fails every time. I have followed the document but, even then i get an error. I get some error related to snappy.c. Below are the logs

Node version = 8.11.1
npm version = 6.0.0
nuget version = 4.6.2 (64 bit)
Installed Visual studio 2013
Open SSL = V 1.0.2 (64bit)

I have followed the installation process correctly, but still unable to build this

C:\Users\myID\git\dummy-kafka-forwarder>npm install --msvs_version=2013

> node-rdkafka@2.3.2 preinstall C:\Users\myID\git\dummy-kafka-forwarder\node_modules\node-rdkafka
> node util/configure

Skipping run because we are on windows

> node-rdkafka@2.3.2 install C:\Users\myID\git\dummy-kafka-forwarder\node_modules\node-rdkafka
> node-gyp rebuild

C:\Users\myID\git\dummy-kafka-forwarder\node_modules\node-rdkafka>if not defined npm_config_node_gyp (node "C:\Users\myID\AppData\Roaming\npm\node_modules\npm\node_modules\npm-lifecycle\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.js" rebuild )  else (node "C:\Users\myID\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\bin\node-gyp.js" rebuild )
Warning: Missing input files:
C:\Users\myID\git\dummy-kafka-forwarder\node_modules\node-rdkafka\deps\librdkafka\win32\outdir\v120\x64\Release\librdkafka.dll
C:\Users\myID\git\dummy-kafka-forwarder\node_modules\node-rdkafka\deps\librdkafka\win32\outdir\v120\x64\Release\librdkafkacpp.dll
C:\Users\myID\git\dummy-kafka-forwarder\node_modules\node-rdkafka\deps\librdkafka\win32\outdir\v120\x64\Release\zlib.dll
Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
  Copying C:\Users\myID\git\dummy-kafka-forwarder\node_modules\node-rdkafka/deps/librdkafka/win32/outdir/v120/x64/Release/librdkafkacpp.dll to C:\Users\myID\git\dummy-kafka-forwarder\node_modules\node-rdk
[...]

PR #493, opened 2018-09-28
Fix Windows build

iradul:
Instead of compiling librdkafka this will download and use precompiled NuGet package.
No need to install whole Microsoft Visual Studio 2013 which is btw removed from Microsoft website.
Required tools are VS Build Tools 2015 and Python 2.7, both easy install-able with [`windows-build-tools`](https://github.com/nodejs/node-gyp#on-windows).

librdkafka depends on `msvcr120.dll` and `msvcp120.dll`.
Unfortunately, `msvcp120.dll` library is missing from both VS Build Tools 2015 and [NuGet librdkafka](https://www.nuget.org/packages/librdkafka.redist/) which is why I bundled it within this fix. File gets extracted into `build/Release/` along with other NuGet librdkafka dlls during installation.

Other option would be installing [VS Build Tools 2013](https://www.microsoft.com/en-us/download/details.aspx?id=40760) since this does include `msvcp120.dll`.

Related issue #413


tsachev:
@iradul I was trying out your PR and it kinda worked for me - 10x a ton. Building on windows is total PITA for me.
One question `msvcr120.dll` was also missing for me during run-time. Do you think it should be added to the zip or it is supposed to be on the target machine?
I have added `node-pre-gyp` to my fork and when the pre-build binaries (produced on our build server) were downloaded on a clean machine (without windows-build-tools) I needed to also add the `msvcr120.dll` manually in order to make it work.

iradul:
@webmakersteve any thoughts on this?

webmakersteve:
I'm a little hesitant to make this change, but I understand the windows community wants it to work like this. Node-gyp and native node pieces in general are pretty bad on Windows, so I'm ok moving to a static binary to link against because Windows is also generally only used for dev.

I think I'm ok to allow it, as long as we change the README to be really clear that windows does NOT compile from source, and requires nuget. And then, we just need to rebase on to the binding changes.

iradul:
I totally understand. My company is a heavy user of Kafka and most of the dev team is using Windows. I've built some neat tools in NodeJS on the top of this library but it's so painful for people to use it on Windows.
Anyway there are two things that I don't like very much here and want to point them out :
* `msvcp120.dll` is bundled within `./deps/windows-install.zip`, now since we're bundling we could also just bundle the whole `librdkafka` binary and forget about NuGet
* every time you upgrade to a newer version of `librdkafka` you need to update first line of `./deps/windows-install.py` (currently it's `librdkafkaVersion = '0.11.5'`), `git describe --tags` might give us autodetection but i didn't want to complicate things further

webmakersteve:
Let's just make it so the package.json has a field for the Librdkafka version, and the script can use node to evaluate that to get the version it should download.

webmakersteve:
Can we not get those DDLs from somewhere else? I'm pretty unfamiliar with the nuget workflow, but including any DDLs inside the codebase seems like it goes against the goal of fetching what we need.

iradul:
I removed zip file with `msvcp120.dll` and updated README by adding *Visual C++ Redistributable Packages for Visual Studio 2013* requirement.
I also added `"librdkafka": "0.11.5",` to `package.json`. This will tell installer which version of librdkafka to download from NuGet.

webmakersteve:
@xapasnon Look to the librdkafka update this week if it is indeed stable. Have some work to do around message headers and admin client this week as well.

edenhill:
librdkafka.redist.1.0.0-RC8 includes the missing msvcp120.dll, please try it out.
https://www.nuget.org/packages/librdkafka.redist/1.0.0-RC8

iradul:
> librdkafka.redist.1.0.0-RC8 includes the missing msvcp120.dll, please try it out.
> https://www.nuget.org/packages/librdkafka.redist/1.0.0-RC8

Works like a charm with the latest `1.0.0`. Thanks @edenhill !

PR #507, opened 2018-10-20
Updated mocha + small code changes for cpplint

mimaison:
- `npm audit` reports security issues with mocha 2.x. I used `npm audit fix --force` to update the `package.json` file automatically.

    However since mocha 4, `process.exit()` is not called automatically and `make e2e` was not terminating once all tests had passed. So it looks like some resources are not currently closed properly. It's something we should fix but in the meantime I've added the `--exit` flag to force the previous behavior (https://boneskull.com/mocha-v4-nears-release/#mochawontforceexit).

- `make lint` also has a number of issues and I fixed the low hanging fruits.

- Minor code/README fixes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ check: node_modules/.dirstamp
 	@$(NODE) util/test-compile.js
 
 e2e: $(E2E_TESTS)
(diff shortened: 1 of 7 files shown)


webmakersteve:
Looks good! Merged.

PR #519, opened 2018-11-06
chore: AdminClient smoke tests

alexander-alvarez:


alexander-alvarez:
@webmakersteve  Here's a simple failing testcase that shows errors aren't being raised properly from the AdminClient.

I've had it happen where I was getting these errors on the server, but no error in the `createTopic` function call.
`org.apache.kafka.common.errors.TopicExistsException: Topic 'testing-123456' already exists.`
```
org.apache.kafka.common.errors.InvalidReplicationFactorException: Replication factor: 3 larger than available brokers: 1.
[2018-11-06 13:32:17,155] INFO [Admin Manager on Broker 1]: Error processing create topic request for topic my-bud with arguments (numPartitions=3, replicationFactor=3, replicasAssignments={}, configs={}) (kafka.server.AdminManager)
```

webmakersteve:
Thanks for the contribution. AdminClient is definitely not done and has not been sufficiently tested. Going to see how far I can get today, but these are helpful!

PR #521, opened 2018-11-08
connection.failure error emit

alexander-alvarez:
Currently emitting `self.metrics` which is an empty object before the connection has been created and gobbling the librdkafka error, so it's pretty useless for figuring out what happened.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -175,7 +175,7 @@ Client.prototype.connect = function(metadataOptions, cb) {
         next(err); return;
       }, 10000).unref();
 
-      self.emit('connection.failure', self.metrics);
+      self.emit('connection.failure', err, self.metrics);
     } else {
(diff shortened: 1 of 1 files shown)


alexander-alvarez:
@webmakersteve doing it which way?

I was thinking of that too, but I was O.K with it because `self.metrics` is an empty object and because of that I think it would have less of an impact; said differently, there's not much you can do with an empty object that you can't do with an Error object.

I would be willing to bet most of the code is written like so
```js
producer.on('connection.failure', (err) => {
  if (err) {
    console.log(err);
  }
});
```
and doesn't even realize it's an empty object

webmakersteve:
Alright, you've convinced me. self.metrics isn't really used right now, as you said. So, let's just merge as is :)

PR #526, opened 2018-11-15
test: Add retry test case

alexander-alvarez:
Basically testing if this sort of behavior is supported. If not what are our alternatives when an error occurs (or no commit occurs) in the 'data' event handler.
@webmakersteve  any thoughts on how to tell the library that the consumption of this message failed, and lets try again?


alexander-alvarez:
I addressed this in userland. My thinking was event emitters have no knowledge of the state of consumers of those events so this wouldn't make sense to implement in this library

Issue #474, opened 2018-08-15
Error: librdkafka.so.1: cannot open shared object file: No such file or directory


F483:
**Environment Information**
 - OS: Ubuntu 18.4 
 - Node Version: 9.0.0 and 10.7.0
 - NPM Version: 5.5.1
 - node-rdkafka version: 2.4.1

**Steps to Reproduce**

```
$ cat package.json 
{
  "name": "test-rdkafak",
  "scripts": {
    "test": "node test.js"
  },
  "devDependencies": {
    "node-rdkafka": "^2.3.4"
  }
}
```

```
$ cat test.js 
var Transform = require('stream').Transform;

const Kafka = require('node-rdkafka');
var stream = Kafka.KafkaConsumer.createReadStream({
  'metadata.broker.list': 'localhost:9092',
  'group.id': 'librd-test',
  'socket.keepalive.enable': true,
  'enable.auto.commit': false
}, {}, {
  topics: 'test',
  waitInterval: 0,
  objectMode: false
});

stream.on('error', function(err) {
  if (err) console.log(err);
  process.exit(1);
});

stream.pipe(process.stdout);

stream.on('error', function(err) {
  console.log(err);
  process.exit(1);
});

stream.consumer.on('event.error', function(err) {
  console.log(err);
})

```
It seems to install, but not sure. See install log in comment below.
```
$ npm install

...


$ npm ls
test-rdkafak@ /home/f483/dev/testkafka
└─┬ node-rdkafka@2.3.4
  ├── bindings@1.3.0
  └── nan@2.10.0

```
The library it says it can't find does seem to have been built, is this a configuration issue?
```

$ find | grep librdkafka.so.1
./node_modules/node-rdkafka/build/deps/librdkafka.so.1
./node_modules/node-rdkafka/deps/librdkafka/src/librdkafka.so.1

$ npm run test

> test-rdkafak@ test /home/f483/dev/testkafka
> node test.js

/home/f483/dev/testkafka/node_modules/bindings/bindings.js:88
        throw e
        ^

Error: librdkafka.so.1: cannot open shared object file: No such file or directory
    at Object.Module._extensions..node (module.js:670:18)
    at Module.load (module.js:560:32)
    at tryModuleLoad (module.js:503:12)
    at Function.Module._load (module.js:495:3)
    at Module.require (module.js:585:17)
    at require (internal/module.js:11:18)
[...]

PR #535, opened 2018-12-06
Fix Bug #474

mike-coolfront:
Fix librdkafka++.so.1 not searching for librdkafka.so.1 correctly by adjusting the librdkafka++.so.1 build configuration to look in its current directory when loading librdkafka.so.1.
This avoids the issue with librdkafka++.so.1 not finding librdkafka.so.1 when not available in a standard library search path as well as avoiding the loading of an unexpected librdkafka.so.1 version which may be present in a standard library search path.

### DIFF
--- a/util/configure.js
+++ b/util/configure.js
@@ -9,6 +9,8 @@ var baseDir = path.resolve(__dirname, '../');
 var releaseDir = path.join(baseDir, 'build', 'deps');
 
 var isWin = /^win/.test(process.platform);
+var isLinux = /linux/.test(process.platform);
+var ENVVARS = ""
 
 // Skip running this if we are running on a windows system
 if (isWin) {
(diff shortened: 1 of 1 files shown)


mike-coolfront:
Hey webmakersteve, 
Notes regarding my research are in issue #474.

With 2.4.1, I was not seeing correct behavior with ubuntu 16 & 18.  I also did not have any success with modifying binding.gyp to correct the issue. 

I'll see about getting sometime to re-validate the behavior with 2.5.0.
If the issue is still present would you have interest in a dockerfile to recreate the scenario for your review?

mike-coolfront:
ok, got some time to revisit and put together a shareable docker test project.
In summary 2.5.x looks good.

2.5.1 - expected lib path is used when librdkafka.so.1 is not on the host system and if there is a preexisting instance
2.5.0 - same as 2.5.1
2.4.2 - fails on both cases using ubuntu 18

not sure if this will be of any value but here is a link to the MIT licensed test project I cobbled together.
[docker based lib path tester for node-rdkafka ](https://github.com/profit-strategies/node-rdkafka-lib-search-tests)

PR #536, opened 2018-12-10
Add higher level producer enrichment wrapper

webmakersteve:
This will support use cases where you want acknowledged produces and guarantees, or need to use the offset for additional functionality after a message is delivered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,35 @@ The following table describes types of events.
 | `event.throttle` | The `event.throttle` event emitted  when `librdkafka` reports throttling. |
 | `delivery-report` | The `delivery-report` event is emitted when a delivery report has been found via polling. <br><br>To use this event, you must set `request.required.acks` to `1` or `-1` in topic configuration and `dr_cb` (or `dr_msg_cb` if you want the report to contain the message payload) to `true` in the `Producer` constructor options. |
 
+### Higher Level Producer
+
+The higher level producer is a variant of the producer which can propagate callbacks to you upon message delivery.
+
+```js
+var producer = new Kafka.HighLevelProducer({
+  'metadata.broker.list': 'localhost:9092',
+});
+```
+
+This will enrich the produce call so it will have a callback to tell you when the message has been delivered. You lose the ability to specify opaque tokens.
+
+```js
+producer.produce(topicName, null, Buffer.from('alliance4ever'), null, Date.now(), (err, offset) => {
+  // The offset if our acknowledgement level allows us to receive delivery offsets
+  console.log(offset);
+});
+```
+
+Additionally you can add serializers to modify the value of a produce for a key or value before it is sent over to Kafka.
+
+```js
+producer.setValueSerializer(function(value) {
+  return Buffer.from(JSON.stringify(value));
+});
+```
+
+Otherwise the behavior of the class should be exactly the same.
+
 ## Kafka.KafkaConsumer
 
 To read messages from Kafka, you use a `KafkaConsumer`.  You instantiate a `KafkaConsumer` object as follows:
--- a/examples/high-level-producer.md
+++ b/examples/high-level-producer.md
@@ -1,0 +1,28 @@
+```js
+var Kafka = require('../');
+
+var producer = new Kafka.HighLevelProducer({
+  'metadata.broker.list': 'localhost:9092',
+});
+
+// Never allow keys
+producer.setKeySerializer(function(v) {
+  return null;
+});
+
+// Take the message field
+producer.setValueSerializer(function(v) {
+  return Buffer.from(v.message);
+});
+
+producer.connect(null, function() {
+  producer.produce('test', null, {
+    message: 'alliance4ever',
+  }, null, Date.now(), function(err, offset) {
+    // The offset if our acknowledgement level allows us to receive delivery offsets
+    setImmediate(function() {
+      producer.disconnect();
+    });
+  });
+});
+```
--- a/examples/high-level-producer.md
+++ b/examples/high-level-producer.md
@@ -1,4 +1,4 @@
-// ```js
+```js
 var Kafka = require('../');
 
 var producer = new Kafka.HighLevelProducer({
@@ -9,9 +9,8 @@ var producer = new Kafka.HighLevelProducer({
 producer.setKeySerializer(function(v) {
   return new Promise((resolve, reject) => {
     setTimeout(() => {
-      console.log('done key serializing');
       resolve(null);
-    }, 1000);
+    }, 20);
   });
 });
 
@@ -24,12 +23,10 @@ producer.connect(null, function() {
   producer.produce('test', null, {
     message: 'alliance4ever',
   }, null, Date.now(), function(err, offset) {
-    console.log(err);
-    console.log(offset);
     // The offset if our acknowledgement level allows us to receive delivery offsets
     setImmediate(function() {
       producer.disconnect();
     });
   });
 });
-// ```
+```
--- a/lib/producer/higher-level-producer.js
+++ b/lib/producer/higher-level-producer.js
@@ -1,0 +1,201 @@
+/*
+ * node-rdkafka - Node.js wrapper for RdKafka C/C++ library
+ *
+ * Copyright (c) 2016 Blizzard Entertainment
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE.txt file for details.
+ */
+
+module.exports = HighLevelProducer;
+
+var util = require('util');
+var Producer = require('../producer');
+var LibrdKafkaError = require('../error');
+var EventEmitter = require('events').EventEmitter;
+var RefCounter = require('../tools/ref-counter');
+var shallowCopy = require('../util').shallowCopy;
+var isObject = require('../util').isObject;
+
+util.inherits(HighLevelProducer, Producer);
+
+function noopSerializer(v) {
+  return v;
+}
+
+/**
+ * Create a serializer
+ *
+ * Method simply wraps a serializer provided by a user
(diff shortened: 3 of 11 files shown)


codeburke:
Why not create a `HighLevelProducer` that extends the `Producer` prototype and then overwrite `produce`? The benefit of that approach is that people (like me) could extend the `HighLevelProducer` with additional hooks like serialization without affecting the `librdkafka`-mapped API of Producer? 

A similar case could be made for a `HighLevelConsumer` that did more auto-magic things like maintaining per TopicPartition highwater offsets to help prevent async processing from committing below the high watermark. 

I'd be more than happy to help out with any of these efforts.

webmakersteve:
I thought about doing it that way, but felt like it was more "opt in" behavior than a separate instance of the class. But I could go either way on this. 

Additionally, I thought to myself, that this would let you make one producer but get two handles: one with enriched functionality (potentially) and the other with the standard one, and allow you to determine which elements you want the more expensive features for. 

Could do the same by providing downgraded methods in the higher level class too though, so it's basically a wash, and I'm fine with whichever way other people would prefer to interface with it.

This discussion is a big reason I wanted this to be a PR and get some baking from the community 😃

codeburke:
Yeah, I understand and could see this get some good conversation. The upgradable/downgradable is just an odd pattern I haven't seen much of in other projects. Since there are assumptions being made on how you instantiate/use the different "levels", it seems to point to multiple classes to KISS. That way you could enforce at instantiation time that the correct options are set instead of creating a high/low-level instance that gets upgraded/downgraded at some point in its life cycle. Obviously pre-req checks can happen in `higherLevel()`... lots of ways to solve this problem.

webmakersteve:
Just updated the PR to use a custom class, and add support for custom serializers on the producer side.

README.md
```diff
@@ -291,6 +291,35 @@ The following table describes types of events.
 | `event.throttle` | The `event.throttle` event emitted  when `librdkafka` reports throttling. |
 | `delivery-report` | The `delivery-report` event is emitted when a delivery report has been found via polling. <br><br>To use this event, you must set `request.required.acks` to `1` or `-1` in topic configuration and `dr_cb` (or `dr_msg_cb` if you want the report to contain the message payload) to `true` in the `Producer` constructor options. |
 
+### Higher Level Producer
+
+The higher level producer is a variant of the producer which can propagate callbacks to you upon message delivery.
+
+```js
+var producer = new Kafka.HighLevelProducer({
+  'metadata.broker.list': 'localhost:9092',
```
codeburke: I'd suggest that we support the `key.serializer` and `value.serializer` config settings that the Java KafkaProducer supports: https://kafka.apache.org/090/javadoc/index.html?org/apache/kafka/clients/producer/KafkaProducer.html

```suggestion
  'metadata.broker.list': 'localhost:9092',
  'key.serializer': NoopSerializer,
  'value.serialzer': NoopSerializer,
```

README.md
```diff
@@ -291,6 +291,35 @@ The following table describes types of events.
...
+producer.produce(topicName, null, Buffer.from('alliance4ever'), null, Date.now(), (err, offset) => {
+  // The offset if our acknowledgement level allows us to receive delivery offsets
+  console.log(offset);
+});
+```
+
+Additionally you can add serializers to modify the value of a produce for a key or value before it is sent over to Kafka.
+
+```js
+producer.setValueSerializer(function(value) {
```
codeburke: If we do support the `key.serializer` and `value.serializer` then should we still support a setter like this?

examples/high-level-producer.md
```diff
@@ -0,0 +1,28 @@
+```js
+var Kafka = require('../');
+
+var producer = new Kafka.HighLevelProducer({
+  'metadata.broker.list': 'localhost:9092',
```
codeburke:
```suggestion
  'metadata.broker.list': 'localhost:9092',
  'key.serializer': NoopSerializer,
  'value.serialzer': NoopSerializer,
```

examples/high-level-producer.md
```diff
@@ -0,0 +1,28 @@
...
+});
+
+// Never allow keys
```
codeburke: Lol, a new definition of "throwing away the keys" 😆

lib/producer/higher-level-producer.js
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Create a serializer
+ *
+ * Method simply wraps a serializer provided by a user
+ * so it adds context to the error
+ *
+ * @returns {function} Serialization function
+ */
+function createSerializer(serializer) {
```
codeburke: Should this support async serializers? When initializing a `Producer` or `HighLevelProducer` you aren't required to pre-define your topics, which means that any serializer that requires some kind of async action would have to happen in the flow of the `produce` call. An example of this would be Confluent's Schema Registry service.

lib/producer/higher-level-producer.js
```diff
@@ -0,0 +1,201 @@
...
+  if (!(this instanceof HighLevelProducer)) {
+    return new HighLevelProducer(conf, topicConf);
+  }
+
+  // Force this to be true for the high level producer
+  conf = shallowCopy(conf);
+  conf.dr_cb = true;
+
+  // producer is an initialized consumer object
+  // @see NodeKafka::Consumer::Init
```
codeburke: typo?

lib/producer/higher-level-producer.js
```diff
@@ -0,0 +1,201 @@
...
+    clearInterval(self._hl.pollingRefTimeout);
+  });
+
+  // Default poll interval. More sophisticated polling is also done in create rule method
+  this.setPollInterval(1000);
+
+  // Listen to all delivery reports to propagate elements with a _message_id to the emitter
+  this.on('delivery-report', function(err, report) {
+    if (report.opaque && report.opaque.__message_id !== undefined) {
+      self._hl.deliveryEmitter.emit(report.opaque.__message_id, err, report.offset);
```
codeburke: Is an event emitter for handling callbacks too heavy weight? Wouldn't it be easier to have have a single callback function that gets called on every `delivery-report` that just does a hash lookup for `__message_id`? Maybe I'm not following why we need an even emitter here.

lib/producer/higher-level-producer.js
```diff
@@ -0,0 +1,201 @@
...
+  this._hl.deliveryEmitter.once(opaque.__message_id, function(err, offset) {
+    self._hl.pollingRef.decrement();
+    setImmediate(function() {
+      // Offset must be greater than or equal to 0 otherwise it is a null offset
+      // Possibly because we have acks off
+      callback(err, offset >= 0 ? offset : null);
+    });
+  });
+
+  return this._oldProduce(topic, partition,
```
codeburke: If we did move to some kind of async `*Serializer` then this logic would have to change to get the serialized values and only produce once the serialization is complete (or blows up)

codeburke:
BTW, thank you for taking the time to explore this option of serialization support in this library. I'm super excited to see this coming (even though it renders the project I was about to publish useless)

lib/producer/higher-level-producer.js
webmakersteve: Sure. I think that makes sense. We can make them support a callback, or potentially do backwards compatible checking for promise methods and utilize those.

lib/producer/higher-level-producer.js
webmakersteve: Indeed it is!

webmakersteve:
@codeburke If you'd like, you can take a stab at the Higher Level Consumer. Initially we can make it support serialization/deserialization, but eventually I'd like to roll the `consumeLoop` (flowing mode) behavior into it as well. 

And thanks for the review!

codeburke:
Sure thing! I'll give the HLC a shot. Do you want me to fork a branch from this branch for review?

FWIW, the way I approached `consumeLoop` in my own extension was overwriting it with a infinite loop of batched `consume` calls to make sure that my app didn't lag-to-death with all the events being read in and emitted into the event buffer. Back-pressure all the way down into the C++ layer would be so nice!

webmakersteve:
I just pushed up my initial stab at async serializer support, which will require promises as it is currently designed. 

Can try to also support callbacks.

webmakersteve:
Sure! Fork this branch, or make a new one. They operate pretty independently. We can merge them in separately if you make the consumer its own branch.

codeburke:
+1 on supporting serialization callbacks to avoid the overhead of Promises from a perf-perspective.

lib/producer/higher-level-producer.js
```diff
@@ -0,0 +1,278 @@
...
+    });
+  });
+
+  var resolvedSerializedValue;
+  var resolvedSerializedKey;
+  var calledBack = false;
+
+  // Actually do the produce with new key and value based on deserialized
+  // results
+  function doProduce(v, k) {
```
codeburke: Since `produce` is now async you can't rely on the return value so it needs to wrap and bubble up errors through the `callback`.

```suggestion
if (typeof callback !== 'function') {
    throw new TypeError('"callback" must be a function when calling HighLevelProducer.producer()');
}

function doProduce(v, k) {
    try {
        self._oldProduce(topic, partition, v, k, timestamp, opaque);
    } catch(err) {
        callback(err);
    }
}
```

webmakersteve:
Added support for callback serializers in addition to promise serializers and unit tests.

codeburke:
I promise I haven't forgotten about this. Been busy at work with the end of year rush, but should have a PR early next week!

AlexKonstantinov1991:
Hello!
I need this kind of functionality in my project, is it safe to use this branch currently? If not, i would like to know whether this feature will be included? And when if it is so?

Issue #538, opened 2018-12-12
AdminClient::CreateTopic only returns the error code and not the included description


dnwe:
**Environment Information**
 - node-rdkafka version: 2.5.0

**Steps to Reproduce**

1. configure kafka with a [KIP-108](https://cwiki.apache.org/confluence/display/KAFKA/KIP-108%3A+Create+Topic+Policy) policy authorizer to return a PolicyViolationException with a descriptive message when a create topic request is received with an unauthorized parameter
2. use the new node-rdkafka 2.5.0 create topic api to submit a create topic request that will be rejected by the policy
3. node-rdkafka returns a error with the generic policy violation error code, but doesn't include the descriptive message from the broker that explains the reason for the rejection:
```
{ Error: Broker: Isolation policy volation
    at Function.createLibrdkafkaError [as create] (/Users/example/node_modules/node-rdkafka/lib/error.js:261:10)
    at /Users/example/node_modules/node-rdkafka/lib/admin.js:132:28
     message: 'Broker: Isolation policy volation',
     code: 44,
     errno: 44,
     origin: 'local' },
```

In the node-rdkafka admin.cc https://github.com/Blizzard/node-rdkafka/blob/d2fde4a3a905976b9333a1f6ce9e2780d78568e6/src/admin.cc#L232-L239 only the generic error code is extracted (via `rd_kafka_topic_result_error`) and returned to the user. However, `rd_kafka_topic_result_error_string(terr)` should also be called to extract and return the accompanying detailed message.

(Related: I raised https://github.com/edenhill/librdkafka/issues/2148 to cover the violation error code string in librdkafka not being quite right either)

PR #539, opened 2018-12-13
Return broker error messages for create topic, create partitions

edoardocomar:
On failing to create a Topic or adding partitions, if the broker returns an error message,
this is now passed back into the JS error.

fixes https://github.com/Blizzard/node-rdkafka/issues/538

note : currently `librdkafka` is returning a fixed error message when create partitions fails,
that will being fixed with a separate PR to `librdkafka` see https://github.com/edenhill/librdkafka/issues/2154

Co-authored-by: Edoardo Comar <ecomar@uk.ibm.com>
Co-authored-by: Mickael Maison <mickael.maison@gmail.com>


webmakersteve:
Looks good to me! Going to merge :)

PR #549, opened 2018-12-25
Changed index.d.ts event emitter keys

KromDaniel:
Hey,
The current event emitter declaration is 

```typescript
 // currently need to overload signature for different callback arities because of https://github.com/Microsoft/TypeScript/issues/15972
    on<T extends keyof EventCallbackRepositoryArityOne, K extends EventCallbackRepositoryArityOne[T]>(val: T, listener: (arg: K) => void): this;
    on<T extends keyof EventCallbackRepositoryArityTwo, K extends EventCallbackRepositoryArityTwo[T]>(val: T, listener: (arg0: K[0], arg1: K[1]) => void): this;
    once<T extends keyof EventCallbackRepositoryArityOne, K extends EventCallbackRepositoryArityOne[T]>(val: T, listener: (arg: K) => void): this;
    once<T extends keyof EventCallbackRepositoryArityTwo, K extends EventCallbackRepositoryArityTwo[T]>(val: T, listener: (arg0: K[0], arg1: K[1]) => void): this;
```

Which is nice but (unfortunately) it doesn't really work in typescript and the overloading did not match the real code, for example events such as `rebalance` that also fires `Error` and `connection.failure'

I changed the `on` and `once` declaration to match the actual code and to be safer (more "`typescript`" style)

-----------------------------

### Before
image

### After
image

image

image

The actual emit args was found from the source code,
if you still see any mismatch please let me know.

**For the horde.**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,12 +20,49 @@ export class Client extends NodeJS.EventEmitter {
 
     offsetsForTimes(topicPartitions: any[], timeout: number, cb?: (err: any, offsets: any) => any): void;
     offsetsForTimes(topicPartitions: any[], cb?: (err: any, offsets: any) => any): void;
+// ON
+    // domain events
+    on(event: 'data', listener: (arg: ConsumerStreamMessage) => void): this;
+    on(event: 'rebalance', listener: (err: Error, assignment: any) => void): this;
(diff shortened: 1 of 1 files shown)


webmakersteve:
> For the horde.

Made me not want to merge it

But I did anyway.

Issue #395, opened 2018-04-06
Dealing with a Producer "memory leak"


fvcaputo:
The last few days I was dealing with a small, but very annoying, memory leak that seemed to start happening when I started using the lib's producer.

I have this process that every few hours it looks for messages to producer on a kafka topic. So the idea here was, create a producer, connect, send, disconnect. It works but I noticed every time a small memory increase that just wouldn't go away.

I believe I have tracked the problem to `setPollInterval`. During the process I mentioned above, I also set an interval to the producer, which I believed was cleared on `disconnect`.  However after reading the documentation and looking at the code I realized that on disconnect the producer does clear the interval, but then it binds a function on `ready` to set it again. The "leak" I was investigating seems to be this listener that was bound on disconnect.

I was able to solve my problem by calling poll during the process, instead of setting the interval. However I kind of feel this looks like a workaround more than anything.

So here I'm really more of posing a question about the behavior of `producer.disconnect()`. I feel that this method should simply clear everything and not have this other side-effect of binding listeners, and instead if someone plans on reconnecting the producer then that person should need to set the interval again. I also feel this is specially bad when using streams, the producer stream sets a poll interval by itself so when you close it you don't even know this is happening internally to the producer instance.

What do you think, @webmakersteve? I could do this change in a small PR if you agree. :)

indieyourself:
@fvcaputo I also meet a memory leak case. Can you show how you fix it?

fvcaputo:
@indieyourself Sorry, we ended up changing the lib to interact with kafka. Currently we are using kafka-node.

webmakersteve:
Disconnect should clear everything. If it doesn't, it's a bug. Would be happy to take a PR for that @fvcaputo !

PR #550, opened 2018-12-26
Remove poll interval on disconnect

fvcaputo:
When disconnecting a producer, if an poll interval was set, there was a new call `self.setPollInterval(interval)`.

But on this new call there is a check `if (!this._isConnected)`. If it is not connected, a new listener will be bound on the "ready" event. This can happen because the flag `_isConnected` is also changed on the event "disconnect". Sometimes, the `setPollInterval` "disconnect" listener will be triggered first, and thus a new function will be bound on "ready".

I believe we should instead clear everything on "disconnect" and not risk biding listeners, so in this change I call `self.setPollInterval(0);` which should clear any interval.

Related discussion: https://github.com/Blizzard/node-rdkafka/issues/395

### DIFF
--- a/lib/producer.js
+++ b/lib/producer.js
@@ -221,9 +221,9 @@ Producer.prototype.setPollInterval = function(interval) {
 
   // Handle disconnections
   this.once('disconnected', function() {
-    // Just rerun this function. It will unset the original
-    // interval and then bind to ready
-    self.setPollInterval(interval);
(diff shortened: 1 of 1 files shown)


fvcaputo:
Seems like the build broke, something with the mac configuration. Not sure if it's something I can change.
@webmakersteve

webmakersteve:
Heh. Weird that I did it like that. Looks good!

PR #554, opened 2019-01-07
Add way to consume messages with a Stream per TopicPartition

JaapRood:
I'd like to propose an alternative approach to consuming messages through a stream, which should not only solve a bunch of outstanding issues, but also increase throughput and performance in certain cases. The approach differs in two significant ways:

- **A separate stream is created for each topic-partition** the consumer is either subscribed to or had manually assigned.
- **A stream is created from an instantiated consumer**, for now proposed as `ReadableStream : consumer.stream(toppar, options)` (but could include returning _all_ streams or a subset of streams).

## Stream per TopicPartition

The current implementation of the _KafkaConsumerStream_ implements a single stream for all the messages that it's consumer might consume. While simple and straightforward, the main issue with becomes apparent when one consumes these messages to do further stream processing with. 

In Kafka, the partition, or more accurately the topic-partition, is the unit of concurrency. Within it the ordering of messages is guaranteed, which makes it an invaluable building block for using the immutable log provided in stream processing. The _ConsumerGroups_, with cluster wide orchestration of which consumer are processing which partitions,  further showcase how partitions are meant to work individually.

When all messages from all topic-partitions come through a single stream, the necessarily have to share back-pressure with each other. Even when splitting it down stream, all partitions share a single buffer. This means that if processing of 1 partition takes a long time, processing for _all_ other partitions have to wait, too. If something goes wrong in this, _all_ partitions have to be stopped. When _ConsumerGroup_ assigns new partitions, _all_ partitions have to be stopped, wait to finished (or canceled) and reconstructed, even if only 1 partition was removed or added.

Instead, when each partition has it's own stream, all of these things can happen individually. If processing from 1 topic is a lot faster than from another one, but they do share the same consumer group (totally reasonable and common!), each partition can be processed at it's own pace.

This approach has a couple of additional benefits, mostly due to the way it allows the message consumption to be implemented (see below for details):

- Rebalance events for consumer groups are only delivered when consuming messages. They _need_ to be reacted to in order for a consumer not to lose it's assigned partitions. A single slowly processed partition can prevent any more consumption to happen. 
- Commit confirmations are delivered as part of consuming messages. But what if you don't want to process (read: consume) the next message until you've confirmed that committing the current one was successful? Unless you get fancy with custom pausing and resuming of partitions so you can call _consumer.consume_ without actually reading any more messages, massively hurting performance, you're in a deadlock! If some other stream causes another read, that won't be an issue.

## Creating streams using `consumer.stream(toppar)`

As highlighted above, using a _ConsumerGroup_, partitions might be re-assigned between various consumers at any time. While the consumer might be long-lived, this might not be true for a stream at all. To stop consuming or processing messages, a Stream should reach it's end, so Node's built-in clean up facilities can be effective. Something has to own and manage these streams across the lifetime of the consumer, making this API a better fit.

While that is one reason for this API ,implementing a stream processing framework, we constantly find that we're referencing _stream.consumer_ in order to perform other operations, like pausing, resuming, seeking, etc. The stream can be a way to consume messages of a consumer, just like consumption per 1 message or through a loop using `consumer.consume()`. That way, it's purely describing the mechanism for _consuming_ messages and doesn't need anything else.

## Implementation thus far and other benefits

In the last couple of months, in a project where we're processing several gigabytes of location events each day, we've implemented our own custom _KafkaConsumerStream_, trying to address some of the issues listed above. While it still used a single stream for all topic partitions, we did uncover some constraints on how to make consumption at _any_ rate performant:

- You want to call `consumer.consume` regularly, as to deliver all the required events and stay known as a healthy worker (not consuming is the metric for determining a node to be unhealthy and in need of termination).
- You can use `consumer.pause` and `consumer.resume` to control for which toppars you want to consume messages by calling `consumer.consume`, while staying assigned to them. This allows processing times for messages that exceed the rebalance response window, delivery reports to be delivered, etc.
- You want to avoid calling `consumer.pause` as it purges the fetch queue `librdkafka` keeps for the toppar in the background. In default configuration, `librdkafka` fetches 100,000 messages per topic partition as quick as it can, ready to be consumed by your application. This creates a buffer that allows for very high throughputs. However, as soon as _pause_ that toppar, the queue is emptied, even if resumed very quickly after. You only want to pause when you're expecting to not read messages for that toppar for a while.

With those constraints in mind, I've taken what I've learned and implemented a proof of concept, with some tests to show that it works. Thus far, the approach generally works as follows:

- The consumer holds a list of all known streams, with a single one for each known topic-partition. 
- A `TopicPartitioionStream` is returned by `consumer.stream(toppar)`, implemented as a [TransformStream](https://nodejs.org/dist/latest-v10.x/docs/api/stream.html#stream_implementing_a_transform_stream) (which is made for streams that can both be written to and read from, where input is logically related to the output).
- As soon as a single stream has been returned the consumer starts consuming messages continuously as soon as the consumer is ready.
- It writes the messages as received to the streams as per their topic and partition. Any stream that indicates to have a full buffer is listed as such.
- At the next read (either after a delay, or not if all buffers streams have room in their buffers) the toppars for streams that are full are paused. Any that have since drained are resumed if they were paused earlier.
- Even if all toppars have been paused, consumption continues with at an interval, making sure we remain alive and receive all necessary events.

--- 

There's a bunch more that would need to happen before this could be merged. 

Most notably, I want to bring over the "throttling" mechanism we developed for the single stream, where we try to adapt our consumption rate (read: waits between `consumer.consume`) calls to match the rate at which they are being read of the stream, to further prevent any pausing of toppars. In our own implementation we used both fetch size + delay, but I haven't figured out yet how to do that for multiple streams, where some toppars could go a lot faster than others.

However, before I put more time into that, I'd like to know whether others agree with the approach and it has a chance of getting merged at all. Please feel free to ask any questions, make comments, etc! 

PS: Please keep in mind that I've tried to get this going as a proof of concept as quick as possible and am aware that the coding style doesn't totally fit in with the rest of the project (mostly in terms of ES2015 / ES2016 features). I'm happy to address all of that, but was hoping to discuss on a more conceptual level first!

### DIFF
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -511,14 +511,14 @@ KafkaConsumer.prototype._consumeStreams = function() {
           // TODO: move this pausing to the start of a consume call, so the stream has a
           // chance to drain before we pause. Remember: avoid all pausing when possible, 
           // as it clears the fetch queue of librdkafka.
-          this.pause([{ topic: stream.topic, partition: stream.partition }])
+          // this.pause([{ topic: stream.topic, partition: stream.partition }])
 
           stream.once('drain', () => {
             // TODO: make this resilient to the streams having ended / removed from the consumer
             // TODO: somehow only resume this if we had to pause it at the last moment
 
             // resume streams now that we have a stream with space in the buffer again
-            this.resume([{ topic: stream.topic, partition: stream.partition }]);
+            // this.resume([{ topic: stream.topic, partition: stream.partition }]);
           })
         }
 
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -128,6 +128,18 @@ function KafkaConsumer(conf, topicConf) {
   this._consumeTimeout = 1000;
   this._waitInterval = 1000;
   this._streams = [];
+  this.on('disconnected', () => {
+    var finishingStreams = this._streams.map((stream) => {
+      const finishing = new Promise((resolve) => stream.once('finish', resolve))
+      stream.end()
+
+      return finishing
+    })
+
+    Promise.all(finishingStreams).then(() => {
+      this.emit('finished')
+    })
+  })
 }
 
 /**
@@ -570,6 +582,14 @@ KafkaConsumer.prototype.stream = function(toppar) {
   const stream = new TopicPartitionStream(toppar);
 
   this._streams.push(stream);
+
+  const removeStream =  () => {
+    this._streams = this._streams.filter((s) => s !== stream)
+  }
+
+  // remove any streams that have ended (either finished or destroyed)
+  stream.once('finish', removeStream);
+  stream.once('close', removeStream);
 
   if (this._streams.length === 1) {
     // start consumption with streams.
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -481,11 +481,6 @@ KafkaConsumer.prototype._consumeNum = function(timeoutMs, numMessages, cb) {
 
 KafkaConsumer.prototype._consumeStreams = function() {
   if (this.consuming) return;
-
-  // TODO: determine fetch size based on available buffer size streams, perhaps timeout too
-  var fetchSize = 1
-  var timeoutMs = this._consumeTimeout || 1000;
-
   // Track whether already in the middle of consuming for streams, so we can make sure the
   // back-pressure is implemented correctly.
   this.consuming = true;
@@ -494,11 +489,12 @@ KafkaConsumer.prototype._consumeStreams = function() {
     return this._streams.find((s) => s.topic === topic && s.partition === partition);
   };
 
+  const readMore = () => {
+    this.consuming = false;
+    this._consumeStreams();
+  }
+
   const readMoreLater = (delay) => {
-    const readMore = () => {
-      this.consuming = false;
-      this._consumeStreams();
-    }
 
     if (delay) {
       setTimeout(readMore, delay).unref();
@@ -506,6 +502,18 @@ KafkaConsumer.prototype._consumeStreams = function() {
       setImmediate(readMore);
     }
   };
+  
+
+  if (!this.isConnected()) {
+    this.once('ready', readMore)
+    return
+  }
+
+  // TODO: determine fetch size based on available buffer size streams, perhaps timeout too
+  var fetchSize = 1
+  var timeoutMs = this._consumeTimeout || 1000;
+
+
 
   const pushMessages = (messages) => {
     return messages.reduce((resume, message) => {
@@ -552,7 +560,7 @@ KafkaConsumer.prototype._consumeStreams = function() {
       // no messages 
       return readMoreLater(this._waitInterval);
     } else {
-      if (this.pushMessages(messages, false)) {
+      if (pushMessages(messages, false)) {
         // TODO: implement the determining of the fetch size and delay for consuming, given
         // the space in buffers of consuming streams. 
         // Use the difference between how much we wanted to fetch and what we actually did
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -607,7 +607,7 @@ KafkaConsumer.prototype._consumeStreams = function() {
   });
 };
 
-KafkaConsumer.prototype.stream = function(toppar) {
+KafkaConsumer.prototype.stream = function(toppar, streamOptions) {
   var toppar = TopicPartition.create(toppar);
   const existing = this._streams.find(({ topic, partition }) => {
     return topic === toppar.topic && partition === toppar.partition;
@@ -615,7 +615,7 @@ KafkaConsumer.prototype.stream = function(toppar) {
 
   if (existing) return existing
 
-  const stream = new TopicPartitionStream(toppar);
+  const stream = new TopicPartitionStream(toppar, streamOptions);
 
   this._streams.push(stream);
 
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -126,6 +126,7 @@ function KafkaConsumer(conf, topicConf) {
   this.topicConfig = topicConf;
 
   this._consumeTimeout = 1000;
+  this._waitInterval = 1000;
   this._streams = [];
 }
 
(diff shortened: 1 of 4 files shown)


JaapRood:
The changes I pushed are to throttle the consumption rate by varying the fetch size. Using a default of `16`, I used the total amount of space available of all streams combined to determine if this needs to be lowered. It's a very basic way to prevent a state where all streams are paused and we have to poll without consuming messages. When the various streams consume at variable rate, it should even do a good job of preventing any pauses at all. However, more advanced mechanics will have to be used to further optimize this. For now, I think this will do while we figure out other details first.

webmakersteve:
I think the idea is definitely sound, and it would be a good feature addition to enable these kinds of use cases to exist. As it stands right now there are a few things I want to mention:

1. This type of consumption isn't a replacement for  the existing consumer stream, which is intended to only make consuming from a kafka consumer easier.
2. I would strongly suggest using a ReadStream rather than a TransformStream, which supports write capabilities. This will also facilitate the backpressure functionalities we need.
3. Streams should probably have their own kafka consumers, and rather than using group based automatically assigned partitioning, should just use manual assignment to ensure they consume from only a single partition. This will avoid the problem of making the c++ consumer the bottleneck for a given stream, and is generally how we can accommodate backpressure appropriately. 
4. I'd like to try our absolute hardest to avoid bringing in highland as a dependency just for this feature. The feature is definitely useful, but probably going to be a more niche case than the general one, and I'd like to spare others from having to have the dependency when they probably won't use it.

Other than that, the code standard guidelines you mentioned will definitely need to be addressed, but I like this feature. 

Hope that's a good enough answer!

JaapRood:
Thanks for your response @webmakersteve. Let me see if I can address your feedback with my thoughts.

First off, I reckon we're talking implementation details in deciding whether to use a `ReadStream` instead of a `TransformStream`, or whether to bring in `highland` or not. So far this has been conceptually challenging enough that I just used those to get something working, have something working and to talk about. Happy to refactor to get that all in it's optimal state, once we know where conceptually this has to go :)

> Streams should probably have their own kafka consumers, and rather than using group based automatically assigned partitioning, should just use manual assignment to ensure they consume from only a single partition. This will avoid the problem of making the c++ consumer the bottleneck for a given stream, and is generally how we can accommodate backpressure appropriately.

This is exactly the constraint I've been running into: in a stream processing environment a _ConsumerGroup_ has to be used for multiple nodes to balance the partitions between them. Initiating a new consumer is not an option, as that would break Kafka's monitoring of how each consumer is performing.

Apart from that constraint, having ran quite a few consumers as part of a single Node.js process, it actually seems to be its Worker Pool that slows things down, up to the point where I've had file I/O take forever, as all resources were being hogged by Kafka clients. Going with the single consumer for an entire group, serving many topic-partitions at once, actually seems to get us higher throughput figures. This might be bottlenecked by how fast the C++ consumer is, but I'd say that's a good spot to be in while processing with Javascript :)

It actually brings us benefits with it as well from a ConsumerGroup point of view, as consuming from multiple partitions makes _consumer.consume_ calls way more likely, keeping us up to date with rebalance events, commit delivery reports, signalling that we're healthy, etc. Where every toppar has it's own Consumer, you quickly end up having to pause the _Consumer_, which is something we'd like to avoid as reasoned at some earlier stage in this PR.

Looking at how streams are used in Node with other things like this, like connections to  other services, sockets, etc, this approach of a small pool of resources seems like a common theme. I don't really have any experiences working on such things, but it's hard for me to think that's a coincidence.

> This type of consumption isn't a replacement for the existing consumer stream, which is intended to only make consuming from a kafka consumer easier.

Putting aside upgrade paths / backwards-compatibility for a second, I'm not sure how this approach isn't exactly that: a super easy way to consume message from Kafka, getting near ideal throughput with low memory usage, etc.

Perhaps, you mean that ending up with heaps of different streams might be difficult, as the user would have to merge them, error handle them separately, etc? If that's the case, I'd suggest we enhance the `consumer.stream()` API to return merged streams already, when either multiple toppars are passed, or all when none have been passed at all?

Interested to hear some more of your thoughts on all of this 😅.

JaapRood:
Activity! This is an important improvement that just takes some time to work out, not something that should be auto closed.

webmakersteve:
I'll need to make adjustments to that thing to make it a lot better, especially since I myself have been so inactive. Will need to add labels or something to make it not go stale.

webmakersteve:
> Apart from that constraint, having ran quite a few consumers as part of a single Node.js process, it actually seems to be its Worker Pool that slows things down, up to the point where I've had file I/O take forever, as all resources were being hogged by Kafka clients. Going with the single consumer for an entire group, serving many topic-partitions at once, actually seems to get us higher throughput figures. This might be bottlenecked by how fast the C++ consumer is, but I'd say that's a good spot to be in while processing with Javascript :)

Have you tried adjusting the number of LIBUV workers in the process? It defaults to something like 4, which is probably way too few when doing work like this. But you're right - creating substreams from a single consumer will reduce the pressure of the consumer thread. I'm fine keeping it like this however because it lends to node's model a bit better, especially since we aren't trying to build a full streaming framework here 😛 

> Putting aside upgrade paths / backwards-compatibility for a second, I'm not sure how this approach isn't exactly that: a super easy way to consume message from Kafka, getting near ideal throughput with low memory usage, etc.

There are times in processing data from Kafka where the partition it originated from is more meaningless to you as a differentiator than it is just a necessary part of doing distributed parallel consumption with concurrent, efficient consumption. In those situations, the consumer stream which generates a fat stream of all the data incoming on the topic is probably going to be a little easier to reason about. 

Merging many streams into a single unit will provide an extra unnecessary step, if this is what you're going for.

However, in situations where the partition is actually meaningful (e.g., hashed partitioning that helps for data locality in caching, or where order is actually significant), this model probably lends itself better to that.

So I'm happy to bring the model into the library. Keeping with the spirit, I think this is what I would probably do:

1. Get the fat stream
2. Write a write stream that has state (a map of substreams) instantiated per partition. Listen to rebalances to delete streams that will no longer be used and are empty in case of partition reassignment. (or don't - this actually introduces a lot of complexity imparted on the user as they need to listen to stream close events and get new streams when partitions are rebalanced).
3. Read incoming data, federate it to streams keyed by partition.
4. Return the map of streams to the user, or a particular stream, if requested.

Does this make sense? I think this gets us the model you're looking for while reducing the amount of code needed and simplifying it a bit.

This also probably means we don't need highland.

If this sounds good to you, let me know, and I'd be happy to merge the change in after a review of the code.

JaapRood:
I think you're on the right path here: as long as each partition can enjoy it's own back-pressure we reach the required amount of isolation to correctly implement Kafka's unit of concurrency. If users don't care about individual partitions and want it to be shared, it's easy to mix them back in together, but the reverse is impossible. 

Unfortunately, though, I can't be the one to implement this anymore. Given the small indie team that we are we had to be pragmatic and switch to a relatively new client written 100% in JS for all new stream processing tasks. Really wish I didn't have to back down from this and possibly cause fragmentation in the already tiny Node + Kafka world :(

PR #556, opened 2019-01-08
Issue-553: Fixes delivery report performance issue

rusty0412:
Fixes issue 553 by batching delivery report in up to 100 deliveries per `DeliveryReportDispatcher::Flush()`. Each `DeliveryReportDispatcher::Flush()` now takes a maximum of about 500 microseconds with an empty callback in a 3GHz core. If there are remaining delivery reports, then Flush() will request a further callback. Test is included in util/test-producer-delivery.js

- Avoid a copy construction of the DeliveryReport in the for loop by using const-references. This, in turn, avoid dynamically allocating the two strings per delivery report. 
- Delivery only requests a callback when the outstanding delivery reports is 1. There is no need to keep requesting the callback if it has already been requested.
- Fixes an issue I encountered where the debug build of node-rdkafka was throwing an exception during configuration in node8, node10, and node11. I could not start the library at all with a debug build. This is a separate commit because it addresses a different issue.

Please let me know of any concerns.

**Tests**
Test verifies that all delivery reports are delivered and in the correct order.
Tested in Linux Node 8 and Mac Node 10. 

Test result before change
```
myhost:node-rdkafka rusty0412$ time node util/test-producer-delivery.js
Producer is ready
Wait took 8 seconds
Finished producing
Reports count match

real	0m12.206s
user	0m13.689s
sys	0m0.904s
```
Test result after change
```
myhost:node-rdkafka rusty0412$ time node util/test-producer-delivery.js
Producer is ready
Wait took 0 seconds
Finished producing
Reports count match

real	0m10.095s
user	0m11.027s
sys	0m0.901s
```

### DIFF
--- a/src/callbacks.cc
+++ b/src/callbacks.cc
@@ -9,6 +9,7 @@
 
 #include <string>
 #include <vector>
+#include <algorithm>
 
 #include "src/callbacks.h"
 #include "src/kafka-consumer.h"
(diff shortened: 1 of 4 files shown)


webmakersteve:
I think there's something wrong with the PR. It looks like the files you changed you were changed in the root directory instead of where the c source files are. Can you fix that and let me know? Then I will re-review!

rusty0412:
You are right, I made this error when I was squashing commits and moving things around for the PR. It is now fixed.

PR #557, opened 2019-01-14
`offset.commit` event documentation and implementation do not match

ankon:
The _event_ only receives the topic partitions, although it will also get called when an error has occurred (see https://github.com/Blizzard/node-rdkafka/blob/56c31c4e81f2a042666160338ad65dc4f8f2d87e/lib/kafka-consumer.js#L93-L104). This PR goes to assume that the implementation is correct and the documentation is wrong, and then fixes the documentation. I guess another way of looking at the same problem is to say the documentation was correct, but the implementation isn't right now, and should be fixed to produce the two arguments (similar to the way the rebalancing is handled). That would introduce a backwards-incompatible change. Let me know if that makes more sense, and I can update this PR accordingly.

While there: the event is missing in `index.d.ts`. That change depends on the above decision though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,7 +358,7 @@ var consumer = new Kafka.KafkaConsumer({
 })
 ```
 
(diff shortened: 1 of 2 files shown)


webmakersteve:
I think this is fine. I'd like to do a pass on the events in general and make them line up a bit better, but that will be for a different day. In the meantime, this is fine :)

PR #579, opened 2019-03-08
Issue 414: Add support for headers

davidtrihy:


Fr8Traindb:
Thanks David!

VijetaH:
+1 for this feature!

josephxsxn:
Having this would help our teams migrate away from some anti-patterns where we use the Kafka Key for header metadata today, it then results in sticky partitions which are consumed unevenly by downstream clients. Moving this metadata into a proper header would let the round robin function when missing a key to work as expected.

R11:
+1 Currently having to work around this limitation is painful for our project

davidtrihy:
@webmakersteve do you think you could take a look at this before the branch goes stale?

iradul:
This works great. Thank you @davidtrihy!
However I think it would be much nicer if headers could look more [NodeJS like](https://nodejs.org/api/http.html#http_response_getheaders). For example instead of this:
```
const headers = [
    { "key1": "v1b" },
    { "key1": "v1a" },
    { "key2": "v2" },
    {
        "key3": "v3",
        "key4": "v4", // this pair will be ignored
    },
]
producer.produce('test_topic', null, Buffer.from('test message'), null, null, null, headers);
```
It would be much nicer to be able to do something like this:
```
const headers = {
    "key1": ["v1b", "v1a"],
    "key2": "v2",
    "key3": "v3",
}
producer.produce('test_topic', null, Buffer.from('test message'), null, null, null, headers);
```
This obvious also apples to the message consumption.

If you guys think this makes sense and you need help implementing this, please let me know.

iradul:
Also worth noticing, this can't be built on Windows because `rdkafkacpp.h` doesn't include appropriate declarations for headers support. It works however with the latest _librdkafka_ `1.0.0` but some work still needs to be done to fix some other Windows build issues. I'll soon make PR to take care of those. That said I think this should be merged only after librdkafka dependecy gets upgraded to `1.0.0` or higher.

davidtrihy:
@iradul thanks for the feedback, I do agree that the way you suggested is cleaner way of doing it, the only problem I have with it is that in C++ code it would be difficult to do this without using something like boost::variant or manipulating the memory in an unsavoury way.

Taking the producer as an example on line 453 of producer.cc when you get the value it's specified that it's a v8String and not an array and I can't immediately see how this would be done easily, however in saying that I'm not hugely familiar with nodejs/nan framework so if you have suggestions I'd be more than happy to hear them.

iradul:
Look at the code that's parsing configuration object, I guess you could do something very similar to that (it's using `v8::Value` and `config_value->IsString()`):
https://github.com/Blizzard/node-rdkafka/blob/da9cbe3fc73f0f134d9cda364114b949e78b4706/src/common.cc#L509-L547

josephxsxn:
Any word on this making its way in? We are going to start using this PR in a Fork which we would prefer not to...

examples/producer.md
```diff
@@ -47,7 +47,10 @@ producer.on('ready', function(arg) {
     var key = "key-"+i;
     // if partition is set to -1, librdkafka will use the default partitioner
     var partition = -1;
-    producer.produce(topicName, partition, value, key);
+    var headers = [
+      { header: "header value" }
+    ]
+    producer.produce(topicName, partition, value, key, new Date(), "". headers);
```
cfinkenstadt: Why `"".` in front of the headers collection?  Looks like this should be `,` instead of `.`.

jibugeorge:
+1. Any update on when this change will be available? We are looking into using this library to produce/consume kafka messages and we need  to be able to use header information.

Fr8Traindb:
I would appreciate an update on this as well.

edenhill:
Two small remarks;
 * Duplicate headers (keys) should be supported without overwriting previous entries.
 * Because of the possibility of duplicates, it is important to preserve the original header order.

tiagoCMatias:
Any update on this?

JaapRood:
> Two small remarks;
> 
> * Duplicate headers (keys) should be supported without overwriting previous entries.
> * Because of the possibility of duplicates, it is important to preserve the original header order.

@edenhill thanks for the info on that. It's probably a good idea to cover that in a test. It pretty much disallows the API of using object of headers straight up (`var headers = { a: '1', b: '' }`, because order is undefined and duplicate keys not allowed), which _is_ something a Node programmer might expect. Should prevent any future PR's from suggesting that without considering that requirement.

jibugeorge:
@davidtrihy Any update on when this feature will be available?

examples/producer.md
webmakersteve: Yeah I think this is a typo here. Would probably be good to get that fixed.

webmakersteve:
Sorry everyone. I should probably post something else somewhere more public on the repo about this, but my time that I am able to have dedicated on this project has reduced substantially. I'm going to try to go through a lot of this today in an effort to catch up :(

webmakersteve:
I'm going to merge this as is, with the README typo. I think the real solution for the object vs array of key value objects is going to be to encapsulate them in a top level type (`Headers`) which can be instantiated `fromObject` or `fromArray`. The produce call can then do checking and convert implicitly if needed.

PR #584, opened 2019-03-22
Add type declarations for HighLevelProducer

macabu:
Adds TypeScript type declarations for the newly added `HighLevelProducer` which were missing.


larslevie:
👍

PR #585, opened 2019-03-22
Update for Windows

terebentina:
On Windows 10 with windows-build-tools I receive
```
..\src\callbacks.cc(275): error C2589: '(': illegal token on right side of '::' [C:\www\htdocs\cust\wt\kafka\apps\packages\node-rdkafka\build\node-librdkafka.vcxproj]
..\src\callbacks.cc(275): error C2059: syntax error: '::' [C:\www\htdocs\cust\wt\kafka\apps\packages\node-rdkafka\build\node-librdkafka.vcxproj]
```
This appears to be due to macros elsewhere redefining `min`. Making the type explicit solves the problem and compiles just fine.

### DIFF
--- a/src/callbacks.cc
+++ b/src/callbacks.cc
@@ -272,7 +272,7 @@ void DeliveryReportDispatcher::Flush() {
   {
     scoped_mutex_lock lock(async_lock);
     outstanding_event_count = events.size();
-    const size_t flush_count = std::min(outstanding_event_count, 100UL);
+    const size_t flush_count = std::min<size_t>(outstanding_event_count, 100UL);
     events_list.reserve(flush_count);
     for (size_t i = 0; i < flush_count; i++) {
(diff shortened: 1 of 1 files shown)


src/callbacks.cc
```diff
@@ -272,7 +272,7 @@ void DeliveryReportDispatcher::Flush() {
   {
     scoped_mutex_lock lock(async_lock);
     outstanding_event_count = events.size();
-    const size_t flush_count = std::min(outstanding_event_count, 100UL);
+    const size_t flush_count = std::min<int>(outstanding_event_count, 100UL);
```
Meisolsson: Should that not be `std::min<size_t>` since that is what they are expecting back?

terebentina: Doh, yes, of course! Fixed and tested.

PR #593, opened 2019-03-28
Emit also the error code in the offset.commit event

edoardocomar:
Without the err information, the emit event is completely useless as one does not know whether it succeeded or failed, and had to use the callback function.

This is a breaking API change, but follows the convention `(err,data)` used for other events.

Co-authored-by: Edoardo Comar <ecomar@uk.ibm.com>
Co-authored-by: Mickael Maison <mickael.maison@gmail.com>

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -327,7 +327,8 @@ describe('Consumer/Producer', function() {
         lastOffset = message.offset;
 
         // disconnect in offset commit callback
-        consumer.on('offset.commit', function(offsets) {
+        consumer.on('offset.commit', function(err, offsets) {
+          t.ifError(err);
           t.equal(typeof offsets, 'object', 'offsets should be returned');
 
           consumer.disconnect(function() {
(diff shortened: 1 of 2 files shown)


edoardocomar:
@webmakersteve just FYI as I discovered that librdkafka retries async commits 
https://gitter.im/edenhill/librdkafka?at=5c9df7181f98a87112bd0daf 
I think having the errcode is a must

edoardocomar:
@webmakersteve bump ...

ankon:
Thanks for pulling this in, it will definitely help us ...

> This is a breaking API change, but follows the convention (err,data) used for other events.

... but it also is a bit dangerous when one actually uses the `offset.commit` event and has something like `"node-rdkafka": "^2.4.0"` in `package.json`, as this change now is in 2.7.0.

webmakersteve:
@ankon Good catch. Sorry I didn't do due diligence there.

This PR adjusts it so the new error-having callback is on a different name: https://github.com/Blizzard/node-rdkafka/pull/629

Also I think future-future, these should actually be raised as a different event, like offsetCommitError or something - I believe that is the general convention for such things.

Issue #591, opened 2019-03-27
node-rdkafka cannot be installed on windows 10


fungiboletus:
**Environment Information**
 - OS: Windows 10
 - Node Version: 10.14.0
 - NPM Version: 6.1.0
 - C++ Toolchain: Visual Studio
 - node-rdkafka version: 2.6.1

**Steps to Reproduce**

`npm install node-rdkafka`

```
node_modules\nan\nan.h(1674): note: see declaration of 'Nan::Callback::Call' (compiling source file ..\src\connection.cc)
..\src\callbacks.cc(275): error C2589: '(': illegal token on right side of '::' [stuff\node_modules\node-rdkafka\build\node-librdkafka.vcxp
roj]
..\src\callbacks.cc(275): error C2059: syntax error: '::' [stuff\node_modules\node-rdkafka\build\node-librdkafka.vcxproj]
```
**Additional context**

It's a bit sad.

fungiboletus:
It's similar to #487 (which is closed but not fixed) with a different error.

terebentina:
@fungiboletus I don't think it's similar to that other issue. For the issue you're having see my PR here: https://github.com/Blizzard/node-rdkafka/pull/585 . It's a really quick fix.

jomaleda:
Do we have any update on this issue? I have problems with version 2.6.1 of node-rdkafka on Windows 10 too.

kowalej-homex:
Same issue over here. I also get a bunch of compilation errors with the new 2.7.0 version. 2.5.1 worked fine.

iradul:
`2.7.1` installs fine on Windows. Please reopen if you're still having issues.

PR #609, opened 2019-04-27
Enhance the travis-ci script

yunnysunny:
I think there are some mistakes about travis-ci :
First, the item of `matrix.include` can only specify one version of node. If multiple given, only the first one will been processed.
Second, the original travis-ci config mixes the test script with deploy script. I think it will be more clear if we separate each other.
Since the reasons mentioned above, I dropped the use of `matrix.include`,  used `jobs.include` instead. There are 4 stages defined in yml file, its names are `test_on_mac` `test_on_win` `create_doc` `publish_num`. The default stage `test` is not specified in yml file, but will also been triggered, and been run on multiple node versions. 
At the end, I add the support of building on windows. Someone said that the building was broken on windows, and now we can test on windows when code pushed. Windows builds are in early access stage in travis-ci , and has some problems, such as the lack of ca files, which may lead to download failed from https link, so I modify some code on python to ignore the ssl verify. Notice that more time will been used when build on windows, I found it takes 4 minutes on travis-ci.
By the way, the readme file tell us to install `windows-build-tools` with vs2015, it may lead to fail for that the node with version greater that 6.x cant not build addon with vs2015.
After testing many times and failed many times, I  finish the enhancement finally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,20 +15,20 @@ language: node_js
 cache:
   directories:
   - node_modules
-node_js:
-  - "4"
-  - "6"
-  - "8"
-  - "10"
+# node_js:
+#   - "4"
+#   - "6"
+#   - "8"
+#   - "10"
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_BUILD_STAGE_NAME" == "Test" ]]; then ./run_docker.sh; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install openssl; fi
   - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then ./win_install.bat; fi
 install:
-  - if [[ "$TRAVIS_BUILD_STAGE_NAME" == "Test" ]]; then npm install; fi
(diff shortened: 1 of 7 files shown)


yunnysunny:
The travis-ci will fail on windows for the issue #591

webmakersteve:
Can you rebase this and comment? It looks good to me, and I merged in the windows fix.

yunnysunny:
I have resolve the conflict in README.md.

webmakersteve:
I'm going to merge this, but it looks like there is a failure on Windows right now. Because of that I don't want those builds to fail the entire CI job so I will be making an adjustment to only warn on windows failures rather than fail the entire build. Once it is more stable I will make it fail the entire build when windows builds fail.

Thanks for the contribution!

yunnysunny:
The current `librdkafka` defined in `package.json` version is `1.0.0.2`, the the version can't been found in `nuget`,  so the build failed in Windows.
I think you have given the wrong version, the newest version on `nuget` is `1.0.1`.

yunnysunny:
I have pulled a new request #634 to resolve it.

PR #618, opened 2019-05-13
Fixing High Level Producer

idangozlan:
1. Added connection check before producing (and throwing an error the same as the default producer).
2. Bug Fix - noopSerializer (was using the native `apply` function instead of the createSerializer one).


JaapRood:
Is this a duplicate of #592?

idangozlan:
It is. Didn't see the previous one, but mine contains also connection check before producing so we rather merge it to solve two issues at the same time

webmakersteve:
I merged the other no op serializer fix first. Can you update this commit after rebasing and adjusting the serializer fix you made?

PR #629, opened 2019-06-04
Fix backwards compatibility problem in node-rdkafka

webmakersteve:
This uses the next level syntax for events which will be camelCase instead of dot separated.

### DIFF
--- a/ci/checks/librdkafka-correct-version.js
+++ b/ci/checks/librdkafka-correct-version.js
@@ -41,21 +41,31 @@ function parseLibrdkafkaVersion(version) {
   const patch = (intRepresentation & patchMask) >> (8 * 1);
   const rev = (intRepresentation & revMask) >> (8 * 0);
 
-  return {
+  const output = {
     major,
     minor,
     patch,
-    rev
   };
+
+  if (rev !== 255) {
+    output.rev = rev;
+  }
+
+  return output;
 }
 
 function versionAsString(version) {
-  return [
+  const versionItems = [
     version.major,
     version.minor,
-    version.patch,
-    version.rev
-  ].join('.');
+    version.patch
+  ];
+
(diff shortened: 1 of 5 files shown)


ankon:
I'm looking how to handle the different versions with this change, and I think we would need some form of "detection" ability:
```typescript
const handleOffsetCommit = (err: Error|undefined, topicPartitions: Array<{topic: string, partition: number, offset: number}>) => {
  // handle stuff
};
// node-rdkafka before 2.7.0, and after 2.7.1: Only topicPartitions are available.
// node-rdkafka 2.7.0: We get the correct signature.
consumer.on('offset.commit', () => {
	if (arguments.length === 2) {
		handleOffsetCommit(arguments[0], arguments[1]);
	} else if (Array.isArray(arguments[0])) {
		handleOffsetCommit(undefined, arguments[0]);
	}
});
consumer.on('offsetCommit', handleOffsetCommit);
```

The problem is that in versions >2.7.0 the handling would now happen twice, once because the `offset.commit` event has only one argument, and once because the `offsetCommit` event exists.

I _think_ for me I can solve the problem by explicitly requiring _exactly_ 2.7.0 now, and as soon as whatever new version is out updating to that version -- right now I the complete landscape of which dependency versions are used in our organisation. But that may not be possible for everyone.

Thoughts?

ankon:
> Thoughts?

One idea that should work for me: Instead of using the `offset.commit`/`offsetCommit` events, I could configure something like `offset_commit_cb: handleOffsetCommit`.

webmakersteve:
I think I'll just leave it as is. Unfortunately a breaking change got through. Hopefully the impact, at this point, would have been unfortunately detected and we just roll forward from now being a bit more mindful.

I do want to eventually refactor the event names to make them camelCased.

PR #630, opened 2019-06-04
Fix typo in test description

ankon:


e2e/consumer.spec.js
```diff
@@ -222,7 +222,7 @@ describe('Consumer', function() {
       t.equal(0, consumer.assignments().length);
     });
 
-    it('should be able to unsusbcribe', function() {
+    it('should be able to unsubcribe', function() {
```
GeoffreyHervet:
```suggestion
    it('should be able to unsubscribe', function() {
```

ankon: That's somewhat embarassing, thanks!

PR #632, opened 2019-06-04
Avoid a memory leak when assigning fails

ankon:
This is based on reading through the code, actually looking for something else.

The memory is allocated when converting the values in NodeAssign, and in the non-error case the ownership is transfered into the consumer's `m_partitions`, where the memory will be reclaimed in the next call to assign. For the error case however the memory isn't freed there, and it is also not freed in the caller.

This commit reshuffles the logic to only do the swapping conditionally on whether things worked out, but always free the list of partitions (either the old ones if things worked, or the non-working ones in the error case).


webmakersteve:
Good catch. Merged!

PR #635, opened 2019-06-18
Allow compatibility with AWS Lambda Layers

darkynt:
From: https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html

The hard-coded paths for the linux linked build cannot be exploited with AWS Lambda's Layers style of building dependencies into a deployable asset.

If one was to create a layer they'd need to install `librdkafka` and set the prefix-path as `./configure --prefix="/opt"` which would result in `/opt/includes/librdkafka` having the header files. 

I've just added this to the array for now so we could potentially consume/publish as a part of an AWS StepFunctions or Lambda heavy workflow.

Happy to discuss - feels like there's a better solution to acquire the includepaths out there somewhere.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -113,7 +113,8 @@
                   "libraries": ["-lrdkafka", "-lrdkafka++"],
                   "include_dirs": [
                     "/usr/include/librdkafka",
-                    "/usr/local/include/librdkafka"
+                    "/usr/local/include/librdkafka",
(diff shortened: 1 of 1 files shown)


darkynt:
Looks like there's an existing issue with the Windows Build here?

webmakersteve:
I don't think it hurts anything to add this, so happy to allow it in.

PR #639, opened 2019-06-24
Update librdkafka to 1.1.0

hugomd:
This PR bumps the `librdkafka` version to 1.1.0, release notes are here: https://github.com/edenhill/librdkafka/releases/tag/v1.1.0

I'm not sure if we just want to bump or perhaps make appropriate changes to match new configuration properties as of 1.1.0.

✌️

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ I am looking for *your* help to make this project even better! If you're interes
 
 The `node-rdkafka` library is a high-performance NodeJS client for [Apache Kafka](http://kafka.apache.org/) that wraps the native  [librdkafka](https://github.com/edenhill/librdkafka) library.  All the complexity of balancing writes across partitions and managing (possibly ever-changing) brokers should be encapsulated in the library.
 
-__This library currently uses `librdkafka` version `0.11.6`.__
+__This library currently uses `librdkafka` version `1.1.0`.__
 
 ## Reference Docs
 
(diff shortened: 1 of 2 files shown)


ankon:
I think https://github.com/Blizzard/node-rdkafka/blob/master/package.json#L5 this also needs updating for Windows (https://github.com/Blizzard/node-rdkafka/blob/master/deps/windows-install.py#L2-L5)

GeoffreyHervet:
@ankon  done in #645

hugomd:
@ankon @GeoffreyHervet I’ve made those changes, although the build still seems to be failing — did I miss something?

deps/windows-install.py
```diff
@@ -3,7 +3,7 @@
 import json
 with open('../package.json') as f:
     librdkafkaVersion = json.load(f)['librdkafka']
-librdkafkaWinSufix = '7' if librdkafkaVersion == '0.11.5' else '';
+librdkafkaWinSufix = '7' if librdkafkaVersion == '1.1.0' else '';
```
GeoffreyHervet: Why ?

GeoffreyHervet: if you revert this diff, CI should be ✅

hugomd: That’s a good question, I assumed the default should be changed regardless. 

I just got off a 12.5 hour flight, so certainly missing a few things! 😅

GeoffreyHervet: `librdkafkaWinSufix` looks to be only for `dllPath` & `libPath` maybe this variable can removed and dllPath&libPath  updated, because we won't upgrade to `0.11.5` 😅

hugomd: @GeoffreyHervet I’ve reverted this diff ✅

awarenet:
Awesome glad this has been picked up! Using libs that rely on this and local npm is painful :)

IdanAdar:
@GeoffreyHervet do we know when will this get merged?  Urgently it seems like node;8.-6-alpine cannot be used with this OpenSSL fix/support.

mkaminski1988:
👍 Would very much like to see this merged

PR #643, opened 2019-06-28
Fix producer documentation

GeoffreyHervet:
Fix a typo mistake in documentation


GeoffreyHervet:
Added 6e66253 due to [src/producer.cc:430](https://github.com/Blizzard/node-rdkafka/blob/master/src/producer.cc#L430) `Nan::ThrowError("Timestamp must be a number");`

PR #645, opened 2019-06-28
Fix CI - librdkafka from nuget

GeoffreyHervet:
According to #627, upgrade in `package.json` _librdkafka_ version.
The bug was librdkafka.redist doesn't exists in version `1.0.0.2` - https://www.nuget.org/packages/librdkafka.redist/


webmakersteve:
Competing PR upgrades to 1.1, which is in the nuget repo, so going to close this one.

PR #657, opened 2019-07-24
Allow Node 8 and above instead of Node 12

IdanAdar:
The engine change restrict package usage to Node 12. This must be reduce to at least Node 8. @webmakersteve


webmakersteve:
Going to adjust this to >= `6`. Was convinced in the PR discussion when someone said it didn't do any hard enforcement of the rule.

webmakersteve:
This has been adjusted. Need to figure out why CI didn't signal this problem even though it tests against multiple node versions.

PR #660, opened 2019-07-29
Update Producer.produce TypeScript signature

bojanrajkovic:
It's missing the headers parameter, which means if you want to send headers using
`node-rdkafka` from TypeScript, you have to any-cast the producer, or extend w/
your own typings.


iradul:
LGTM

Issue #648, opened 2019-07-04
HighLevelProducer: Support custom headers


alkiskal:
Is there a reason that the HighLevelProducer doesn't support custom headers?

bojanrajkovic:
I was about to hack this in, but it would be a breaking API change, unless you add the headers parameter _after_ the callback, which would look weird. :-/

alkiskal:
Yes... You could track the arguments length and come up with a hackish way to handle the old invocation as well, but I wouldn't go this way. Another way would be to make the last argument either the callback or an object, which will contain the callback and the headers. In general there is no best way at this point. But being able to set custom headers is important IMHO.

priyabratap:
@alkiskal  I am looking for the same, to pass custom headers. Sorry I did not completely understand your comments. Could you please give some example with sample code?

PR #667, opened 2019-07-31
Add headers support on HighLevelProducer

iradul:
This will add headers support on `HighLevelProducer` #648


ppinel:
This is a breaking change !

GeoffreyHervet:
you should switch callback & header params order `HighLevelProducer.prototype._modifiedProduce = function(topic, partition, message, key, timestamp, callback, headers)` @iradul

Issue #602, opened 2019-04-16
CreateTopic and DeleteTopic in admin.cc do not use the timeout_ms parameter


juddg:
Ubuntu 16.04
node-rdkafka 2.5.1
node 9.11.2

The underlying CreateTopic and DeleteTopic functions in admin.cc do not use the timeout_ms parameter. See [CreateTopic](https://github.com/Blizzard/node-rdkafka/blob/master/src/admin.cc#L177) and [DeleteTopic](https://github.com/Blizzard/node-rdkafka/blob/master/src/admin.cc#L250).
Thus it is not possible to increase the timeout by passing a timeout value to the node-rdkafka createTopic and deleteTopic functions.

PR #668, opened 2019-07-31
Respect timeout_ms param in admin APIs

iradul:
Admin APIs do not respect `timeout_ms` parameter as reported here #602. This will fix it.

### DIFF
--- a/src/admin.cc
+++ b/src/admin.cc
@@ -155,6 +155,7 @@ rd_kafka_event_t* PollForEvent(
   int timeout_ms) {
   // Establish what attempt we are on
   int attempt = 0;
+  int attempt_timeout_ms = max_tries > 0 ? timeout_ms / max_tries : timeout_ms;
 
   rd_kafka_event_t * event_response;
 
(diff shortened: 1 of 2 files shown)


src/admin.cc
```diff
@@ -155,14 +155,15 @@ rd_kafka_event_t* PollForEvent(
   int timeout_ms) {
   // Establish what attempt we are on
   int attempt = 0;
+  int attempt_timeout_ms = max_tries > 0 ? timeout_ms / max_tries : timeout_ms;
```
iradul: @webmakersteve 
All calls to this method will make 5 attempts. The original timeout was `1000ms` but these 5 attempts will effectively turn it into `5000ms` timeout.
Here we'll ensure that timeout is fixed no matter how many attempts there are (it's not exactly fixed for example when `timeout_ms / max_tries` division ends up with some non-zero modulo but that's likely irrelevant deviation)
This might be considered as a breaking change because the default timeout was effectively `5000ms` although it was set to `1000ms`. This is why I increased this default value to 5000 in `lib/admin.js`

benjaminroux:
Looks like we had the same idea
https://github.com/Blizzard/node-rdkafka/pull/653

iradul:
Oh I see, I'm closing this one.

PR #696, opened 2019-10-15
Add TypeScript defs for Producer/Consumer config

ntgn81:
#### This adds a interface definitions for global and topic configurations.

It utilizes a script that
- pulls content from librdkafka/CONFIGURATION.md
- converts that to TypeScript interfaces

A few interesting things that deviate from librdkafka/CONFIGURATION.md
- `dr_msg_cb` - becomes `'boolean'
- `dr_cb` - becomes `'boolean | Function'
- `rebalance_cb` - becomes `'boolean | Function'
- `offset_commit_cb` - becomes `'boolean | Function'

#### Notes
- I'm using 4 spaces as opposed to the 2 space guideline to match with the existing spacing in this file
- The script that pulls from librdkafka documentation - https://gist.github.com/ntgn81/066c2c8ec5b4238f85d1e9168a04e3fb. I could put it as part of this repo, if that is okay

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -79,7 +79,7 @@ export class Client extends NodeJS.EventEmitter {
 export type ErrorWrap<T> = boolean | T;
 
 export class KafkaConsumer extends Client {
-    constructor(conf: any, topicConf: any);
+    constructor(conf: ConsumerGlobalConfig, topicConf: ConsumerTopicConfig);
 
     assign(assignments: any): this;
 
@@ -124,7 +124,7 @@ export class KafkaConsumer extends Client {
 }
 
 export class Producer extends Client {
-    constructor(conf: any, topicConf?: any);
+    constructor(conf: ProducerGlobalConfig, topicConf?: ProducerTopicConfig);
 
     flush(timeout: any, callback: any): any;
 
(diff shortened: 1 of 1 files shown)


nam-asapp:
Not sure why it is failing the CI build. The error don't seem to be related :(

ntgn81:
I'm not sure why it's failing, the error on CI builds don't seem to be related :(

iradul:
This is great!

It would be nice to decouple these auto-generated declarations from the manually created ones.

I have a few suggestions on how to accomplish this:
* instead of `declare` use `export` in generator script
* please add generator script to the repo, you can call it `config.d.ts.generator.js` for example (we certainly want to reuse it in future)
* save generated definitions into `config.d.ts` and make a note in the header that it's generated by `config.d.ts.generator.js`
* in `index.d.ts` import `config.d.ts` interfaces and apply them as you already did but also export them, so for instance:
```index.ts.d
import { Readable, Writable } from 'stream';
import {
    GlobalConfig,
    TopicConfig,
    ConsumerGlobalConfig,
    ConsumerTopicConfig,
    ProducerGlobalConfig,
    ProducerTopicConfig,
} from './config';

export * from './config';

...
```

iradul:
I'll make requested changes my self but I'm still merging this because you must get credit for the great work you did.

iradul:
Also I mentioned you [here](https://github.com/Blizzard/node-rdkafka/blob/master/config.d.ts#L2) (and [here](https://github.com/Blizzard/node-rdkafka/blob/master/ci/librdkafka-defs-generator.js#L125)).

ntgn81:
Yay, thanks for picking up my slack ❤️

PR #718, opened 2019-11-23
expose isDisconnecting flag

keith-chew:


codeburke:
Since `isConnected()` already exists and you are proposing `isDisconnecting()`, my suggestion would be that we create a new enum of ConnectionState values and a new function `getConnectionState()` to return the corresponding value.

```
RdKafka.CODES.CONNECTION_STATE = {
    DISCONNECTED: -2,
    DISCONNECTING: -1,
    CONNECTING: 0,
    CONNECTED: 1
};

// This is super ugly and we should probably update the Client to keep a single variable for
// consistent state that can be returned here.
Client.prototype.getConnectionState = function() {
    if (this._isConnected) {
        return RdKafka.CODES.CONNECTION_STATE.CONNECTED;
    } else if (this._isConnecting) {
        return RdKafka.CODES.CONNECTION_STATE.CONNECTING;
    } else if (this._isDisconnecting) {
        return RdKafka.CODES.CONNECTION_STATE.DISCONNECTING;
    }

    return RdKafka.CODES.CONNECTION_STATE.DISCONNECTED;
}
```

This would allow for truthy checking if a Client was connected or not based on the return value.
```
if (producer.getConnectionState()) {
    // is connected
} else {
   // is not actively connected, could be disconnected/disconnecting/connecting
}
```

codeburke:
@iradul Would love your thoughts on my suggested approach here.

iradul:
@codeburke looks good to me.

keith-chew:
please note that this PR has been included in PR 728:

https://github.com/Blizzard/node-rdkafka/pull/728

which fixes a critical bug for us.

Issue #724, opened 2019-12-05
Producer class type declaration missing attributes


chris--young:
It looks like the typescript declaration for the Producer class is missing some attributes. For example, the createWriteStream method ([defined here](https://github.com/Blizzard/node-rdkafka/blob/master/lib/producer.js#L154)) is exlcuded in the declaration ([here](https://github.com/Blizzard/node-rdkafka/blob/master/index.d.ts#L126)).

**Steps to Reproduce**
1. Try using `Producer.createWriteStream` method
```
import { Producer } from 'node-rdkafka'
Producer.createWriteStream
```
2. Compile
3. Observer error
```
error TS2339: Property 'createWriteStream' does not exist on type 'typeof Producer'.
```

**Additional context**

I'm happy to make some updates to the type declaration, unless this was done intentionally. 

I do see that the `createWriteStream` is its own export so getting around this is as simple as...
```
import { createWriteStream } from 'node-rdkafka'
createWriteStream
```
However I can still imagine people would want the option to do both.

codeburke:
Thanks for the submitting the Issue. Since `node-rdkafka` exposes `createWriteStream` directly, is there a reason you aren't using that?

chris--young:
No, we ended up just using the exposed `createWriteStream` to get around this. I opened the issue anyway cause it still seems like the type definition is wrong. It's very easy thing for me to fix.

codeburke:
Glad the exposed method worked for you. Yeah, I don't see any harm in updating the definition file if you feel like putting in a quick PR to fix that.

PR #726, opened 2019-12-11
Add type for Producer.createWriteStream

chris--young:
Update TypeScript type declaration file to include definiton of
Prouder.createWriteStream method.

Fixes: https://github.com/Blizzard/node-rdkafka/issues/724


chris--young:
I set everything to `any` cause that seems to be the theme of this file. It kind of defeats the purpose of using TypeScript to have it this way though. This whole file really needs a one over to add the proper types.

PR #727, opened 2019-12-11
Total Timeout

keith-chew:
This PR:
- exposes the isDisconnecting flag so the application can manage the lifecycle of the library
- introduces a new  option called Consume Total Timeout, default value is 1000ms
- removes prepack script so we can actually include branches from package.json


keith-chew:
Fixed cpplint errors and replaced PR with:
https://github.com/Blizzard/node-rdkafka/pull/728

PR #728, opened 2019-12-11
Fix/cpplint tidy up

keith-chew:
This PR:
- exposes the isDisconnecting flag so the application can manage the lifecycle of the library
- introduces a new option called Consume Total Timeout, default value is 1000ms
- removes prepack script so we can actually include branches from package.json

### DIFF
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -125,6 +125,7 @@ function KafkaConsumer(conf, topicConf) {
   this.topicConfig = topicConf;
 
   this._consumeTimeout = 1000;
+  this._consumeTotalTimeout = 1000;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


lib/kafka-consumer.js
```diff
@@ -125,16 +125,25 @@ function KafkaConsumer(conf, topicConf) {
...
+  this._consumeTotalTimeout = 1000;
 }
 
 /**
  * Set the default consume timeout provided to c++land
  * @param {number} timeoutMs - number of milliseconds to wait for a message to be fetched
  */
 KafkaConsumer.prototype.setDefaultConsumeTimeout = function(timeoutMs) {
-  this._consumeTimeout = timeoutMs;
+    this._consumeTimeout = timeoutMs;
```
sam-github: random whitespace change?

src/workers.cc
```diff
@@ -539,7 +578,8 @@ void KafkaConsumerConsumeNum::Execute() {
       case RdKafka::ERR__TIMED_OUT:
       case RdKafka::ERR__TIMED_OUT_QUEUE:
         // Break of the loop if we timed out
-        looping = false;
+        // No, rely on total timeout check above
+        // looping = false;
```
sam-github: delete code, instead of commenting it out.

src/workers.cc
```diff
@@ -512,22 +512,61 @@ void KafkaConsumerConsumeLoop::HandleErrorCallback() {
...
+                                     const int & total_timeout_ms) :
   ErrorAwareWorker(callback),
   m_consumer(consumer),
   m_num_messages(num_messages),
-  m_timeout_ms(timeout_ms) {}
+  m_timeout_ms(timeout_ms),
+  m_total_timeout_ms(total_timeout_ms) {}
 
 KafkaConsumerConsumeNum::~KafkaConsumerConsumeNum() {}
+#ifndef _WIN32
```
sam-github: this is used in 3, at least, places, should probably be a function

sam-github:
Btw, changing a branch shouldn't cause a PR to be closed, the branch can just be pushed, or in this branches case, rebased, squashed, and force-pushed.

PR #740, opened 2020-01-08
Add support for python 2 and 3 when installing on Windows

mdowds:
When installing node-rdkafka on Windows, if `python` on your PATH points to python 3, the installation fails with an error about being unable to import urllib2. This PR changes the installation script so it tries to import the python 3 module first, and fall back on the python 2 module. This will allow installation regardless of the python version you have as your default.

### DIFF
--- a/deps/windows-install.py
+++ b/deps/windows-install.py
@@ -18,9 +18,16 @@ libPath = outputDir + '/build/native/lib/win{}/x64/win{}-x64-Release/v120'.forma
 includePath = outputDir + '/build/native/include/librdkafka'
 
 # download librdkafka from nuget
-import urllib2, ssl
+try:
+    # For Python 3.0 and later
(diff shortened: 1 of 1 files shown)


iradul:
There is already PR #706 that is almost identical to this.

alex3165:
@iradul I would get this one merged since there is no need to duplicate `    filedata = urllib2.urlopen(librdkafkaNugetUrl, context=ssl._create_unverified_context())` in try and catch and the commit history on https://github.com/Blizzard/node-rdkafka/pull/706 looks weird. Is it ok if i get this merged?

mmacai:
Would it be possible to merge this change? Right now, it's not possible to develop on Windows since Python 2.x is gone.

iradul:
Thank you @mdowds

PR #745, opened 2020-01-17
add createReadStream to KafkaConsumer type def

jpdstan:
Adds this missing definition from the `KafkaConsumer` API: https://github.com/Blizzard/node-rdkafka/blob/50ca65778c2295a690683857aa84522639be87a2/lib/kafka-consumer.js#L158

PR #746, opened 2020-01-22
index.d.ts broken

f3nrir92:
Package wont be able to compile with ts due to changes in DefinitelyTyped/DefinitelyTyped#40927

Up version after merge pls. This is very critical.
-------
Fix error with EventEmitter caused by DefinitelyTyped/DefinitelyTyped#40927
Add type for KafkaConsumer.createReadStream
Fix return type for Producer.createWriteStream

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,7 @@
 import { Readable, Writable } from 'stream';
+import { EventEmitter } from 'events';
 
-export class Client extends NodeJS.EventEmitter {
+export class Client extends EventEmitter {
     constructor(globalConf: any, SubClientType: any, topicConf: any);
 
     connect(metadataOptions?: any, cb?: (err: any, data: any) => any): this;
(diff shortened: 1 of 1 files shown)


f3nrir92:
Maybe yes if implementation is fully on Typescript. But this is only definitions for type check at compile time.

Now compilation of Typescript project is broken due to changes in `DefinetlyTyped`
```
node_modules/node-rdkafka/index.d.ts:5:29 - error TS2689: Cannot extend an interface 'NodeJS.EventEmitter'. Did you mean 'implements'?

5 export class Client extends NodeJS.EventEmitter {
```

iradul:
The issue happens with `@types/node` version `13`. Node `13` is not officially supported with this library but since the fix is not making any breaking change it's legit.

jpdstan:
`createReadStream` is a `static` property on `KafkaConsumer` - I think [this line](https://github.com/Blizzard/node-rdkafka/pull/746/files#diff-b52768974e6bc0faccb7d4b75b162c99R125) is wrong. here's a PR I submitted before with the correct definition: https://github.com/Blizzard/node-rdkafka/pull/745/files

PR #754, opened 2020-02-17
DOCS: Added .setPollInterval() to Standard Producer API example

nandanrao:
This would have saved me a lot of pain ;)


iradul:
LGTM

Issue #733, opened 2019-12-18
Error Make failed with exit code 2, Node 13


deemalpatel:
**Environment Information**
 - OS [Mac Mojave 10.14.6]: 
 - Node Version [13.2.0]:
 - NPM Version [6.13.1]:
 - node-rdkafka version [2.4.2]:

**Steps to Reproduce**

I am trying install rdkafka@2.4.2 on my project and every time it breaks with 'Error Make failed with exit code 2.' I already have openssl installed via brew and I added the LDFLAGS and CPPFLAGS to my bash profile already but I still am generating this. Anyone have this same issue and was able to solve it? 

**node-rdkafka Configuration Settings**

**Additional context**

gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:194:23)
gyp ERR! stack     at ChildProcess.emit (events.js:210:5)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:270:12)
gyp ERR! System Darwin 18.7.0
gyp ERR! command "/usr/local/Cellar/node/13.2.0/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /Users/cn193394/Click/click-taskvalidator/node_modules/node-rdkafka
gyp ERR! node -v v13.2.0
gyp ERR! node-gyp -v v5.0.5
gyp ERR! not ok 
npm WARN click-taskvalidator@1.0.4 No repository field.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! node-rdkafka@2.4.2 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the node-rdkafka@2.4.2 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:

analytik:
node-rdkafka@2.4.2 is a very old version, it will never build. Your best bet is using Node.js v12 and node-rdkafka@2.7.x - it's unlikely it will build on Node.js v13, but feel free to try. If it doesn't help, you'll need to submit a pull request to add Node.js v13 compatibility, or wait for someone else does.

ntkumar1:
[...]

Issue #628, opened 2019-06-04
Looking for more Collaborators!


webmakersteve:
Hi everybody.

First off, I just want to thank the community for all of the support this repository has gotten. A few years ago when I started working on what would become `node-rdkafka`, I never could have imagined it getting this big! It wouldn't be nearly as big without all of you - your contributions, your bug reports, and the inspiration to keep me working on it. 

This project originated on Blizzard's data team. We built out kafka support in node.js because we were using node for message streaming from HTTP to kafka to be processed by our downstream services. A few months after that, we stopped using node for new projects so we started using the library less. I still supported it as best I could, and we had still had existing services that used it that were made better with every release!

About 6 months ago I moved to another team that doesn't use any node.js, and the amount of time I've had to commit on this repository has gone to its lowest ever. This would be more OK, but my free weekend-time has also been at an all time low.

So, this has all been a long way of saying I need help! I'd like to give some of the responsibility for the management of this project to some of you. If you are interested at all please email me at sparente@blizzard.com! 

I'm hoping to get more active again as best I can to give this project the support it deserves. There's a lot of work to be done on it to make it even better! If that excites you, send me an email!

Thank you everyone for your continued patience. Its totally undeserved, but appreciated :)

JaapRood:
I think I speak on behalf of a lot of people to say: 🎉thank you for all your effort so far and in the future❤️! So many people, including myself personally and my team professionally, have benefited from all of your hard work, getting this setup in the first place.

I really hope someone with experience in both C++ and Node is willing to take on a contributor role, as especially for high-throughput cases it's hard to beat binding with the excellent `librdkafka`. Are there any companies / teams out there (apart from Blizzard) with these skillsets that use `node-rdkafka` to run any production services?

Unfortunately, that excludes myself, as all my C++ experience is just reading and hacking some other people's code. We're only a 2 person indie team, so lacking the resources to address the issues the project has been facing as of late. It pushed us towards adopting a client written 100% in JS for all new stream processing tasks, but without any of the luxuries of `librdkafka`'s throughput, streams, near-mirrored consumer API's, etc.

rusty0412:
What kind of help do you need? Do you have a roadmap of things to improve, modernize? Do you know of bottlenecks that need to be optimized? Would you be open to dropping support for unsupported node releases (<8.0)?

webmakersteve:
@rusty0412 

[...]

PR #759, opened 2020-03-06
Support Node.js 13.x 

sam-github:
Builds and passes tests on Ubuntu with 13.x.                                     
                                                                                 
Also fixes many deprecations.  A few outstanding deprecations related to         
integrating with the new async context tracking API remain, so                   
continuation-local-storage may (still) not work (have not checked).              
                                                                                 
Fixes: https://github.com/Blizzard/node-rdkafka/issues/733 

-----

This was a fairly mechanical fix, I'm not that familiar with the code or usage.

I noticed that `npm test` is passing but `make`  is not, which concerned me, but it turns out the same is true for master. So, I'm still concerned :-), but it isn't a problem intoduced by this PR AFAICT.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -64,24 +64,30 @@ void ConstantsInit(v8::Local<v8::Object> exports) {
   NODE_DEFINE_CONSTANT(topicConstants, RdKafka::Topic::OFFSET_STORED);
   NODE_DEFINE_CONSTANT(topicConstants, RdKafka::Topic::OFFSET_INVALID);
 
-  exports->Set(Nan::New("topic").ToLocalChecked(), topicConstants);
+  Nan::Set(exports, Nan::New("topic").ToLocalChecked(), topicConstants);
 
-  exports->Set(Nan::New("err2str").ToLocalChecked(),
+  Nan::Set(exports, Nan::New("err2str").ToLocalChecked(),
     Nan::GetFunction(Nan::New<v8::FunctionTemplate>(NodeRdKafkaErr2Str)).ToLocalChecked());  // NOLINT
 
-  exports->Set(Nan::New("features").ToLocalChecked(),
+  Nan::Set(exports, Nan::New("features").ToLocalChecked(),
     Nan::GetFunction(Nan::New<v8::FunctionTemplate>(NodeRdKafkaBuildInFeatures)).ToLocalChecked());  // NOLINT
 }
 
-void Init(v8::Local<v8::Object> exports, v8::Local<v8::Object> module) {
+void Init(v8::Local<v8::Object> exports, v8::Local<v8::Value> m_, void* v_) {
+#ifdef NODE_MAJOR_VERSION <= 8
   AtExit(RdKafkaCleanup);
+#else
+  v8::Local<v8::Context> context = Nan::GetCurrentContext();
+  node::Environment* env = node::GetCurrentEnvironment(context);
+  AtExit(env, RdKafkaCleanup, NULL);
+#endif
   KafkaConsumer::Init(exports);
   Producer::Init(exports);
   AdminClient::Init(exports);
   Topic::Init(exports);
   ConstantsInit(exports);
 
-  exports->Set(Nan::New("librdkafkaVersion").ToLocalChecked(),
+  Nan::Set(exports, Nan::New("librdkafkaVersion").ToLocalChecked(),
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "node-rdkafka",
-  "version": "v2.7.4",
+  "version": "v3.0.0",
```
sam-github: I bumped the package major because this is a breaking change, but I can revert it that's not how its done here.

sam-github:
@iradul and perhaps @davidtrihy since you have C++ experience according to https://github.com/Blizzard/node-rdkafka/issues/628#issuecomment-519068981

iradul:
Thank you for this PR!

package.json
iradul: Is this breaking change because versions 8 and older are dropped or is there some API breaking change?

sam-github: breaking because it drops support 8.x and older, no API change

sam-github:
I'll figure out the conflicts tomorrow and rebase.

src/binding.cc
```diff
@@ -142,24 +142,26 @@ void ConstantsInit(v8::Local<v8::Object> exports) {
...
+  Nan::Set(exports, Nan::New("features").ToLocalChecked(),
     Nan::GetFunction(Nan::New<v8::FunctionTemplate>(NodeRdKafkaBuildInFeatures)).ToLocalChecked());  // NOLINT
 }
 
-void Init(v8::Local<v8::Object> exports, v8::Local<v8::Object> module) {
-  AtExit(RdKafkaCleanup);
+void Init(v8::Local<v8::Object> exports, v8::Local<v8::Value> m_, void* v_) {
+  v8::Local<v8::Context> context = Nan::GetCurrentContext();
+  node::Environment* env = node::GetCurrentEnvironment(context);
+  AtExit(env, RdKafkaCleanup, NULL);
```
iradul: This seems to be reason why it doesn't work for node <= 8
This may do the trick:
```
  #if NODE_MAJOR_VERSION <= 8
  AtExit(RdKafkaCleanup);
  #else
  AtExit(node::GetCurrentEnvironment(Nan::GetCurrentContext()), RdKafkaCleanup, NULL);
  #endif
```

skairali:
+1

sam-github:
Builds on `<8.x` now.

src/common.cc
```diff
@@ -534,7 +545,7 @@ rd_kafka_NewTopic_t* FromV8TopicObject(
           err = rd_kafka_NewTopic_set_config(
             new_topic, pKeyString.c_str(), pValString.c_str());
 
-          if (err != RdKafka::ERR_NO_ERROR) {
+          if (err != RD_KAFKA_RESP_ERR_NO_ERROR) {
```
iradul: What's the reason for doing this?

sam-github: Type mismatch. `err` is type `rd_kafka_resp_err_t`, `RdKafka::ERR_NO_ERROR` is of type `enum ErrorCode`, they are numerically equivalent, but not comparable (without warning).

csantanapr:
Thank you for this PR 👍

iradul:
Some e2e tests are failing on v13. I'll have investigate this.

sam-github:
@iradul I didn't find info on how to run the e2e tests, I'll take a shot at fixing issues if I know how to reproduce.

IdanAdar:
@iradul can you help out here?

iradul:
> @iradul I didn't find info on how to run the e2e tests, I'll take a shot at fixing issues if I know how to reproduce.

You first need to start local Kafka, you can use provided `docker-compose -d up` for that. Also, you may need to run it twice or so until Kafka is actually up. This is because sometimes Zookeeper boots slowly and then Kafka fails to load.

After you have Kafka running on `localhost:9092`, you just do `make e2e` to run the tests.

iradul:
I'll merge this since the e2e test issue (an empty string Buffer becomes `null`) is related to a change in v13 NodeJS/V8 code base and how it handles zero length Buffer. It's reported [here](https://github.com/nodejs/node/issues/32089).
I have a workaround that I'll commit soon.

sam-github:
Thanks for looking at this @iradul

ntkumar1:
@iradul When can we expect next version of rdkafka with node.js 13 support

PR #768, opened 2020-03-20
InternalAdminClient should contain connect()

cwadding:
When using typescipt the connect() method on the AdminClient is not accessible since it is not included on the interface. However, it is included in the implementation as shown in https://github.com/Blizzard/node-rdkafka/blob/master/lib/admin.js#L92.


iradul:
AdminClient gets auto-connected on its creation so calling `connect` method is discouraged. Still this is legit but before I can merge this please remove `package.json` version update.

Issue #709, opened 2019-11-14
Consumer has 500ms delay


yunnysunny:
The problem make me annoying these days. We used consumer to consume data from kafka, and found its had 500ms delay at  fixed-frequency.
We checked all the configuration of librdkafka, and found nothing can reduce the delay. And then I add the event listener of `event.log`, and print the partition and offset of current received message to console. Then I found that the fetch event was almost 500ms before the consumer's `data` on same partition and offset.
There must be something  prevent the consumer from returning the data received at once! So I check the code of node c++ and c. Finally, I found a piece of suspicious code:
https://github.com/Blizzard/node-rdkafka/blob/8d2963bf21ed9b8a79ad5adabcec2362bc12392e/src/workers.cc#L450-L459

There is a magic number in the code, and it not indicated on the document, and not supply an option argument for developer to change it.

Sleep function would not make current thread busy, even you call it with 1ms, I think 500ms is not a optimal value as default.

Tapppi:
Looks like the loop version of consume needs a similar optimisation to https://github.com/Blizzard/node-rdkafka/commit/e82bb2ea93d03e56ed9f1f1a1d2fdfaf22220d00#diff-d3b92fea9df0e70b4f0aed1821f7b6c9 . It is indeed likely that you are producing the message at such a point that the consumer is sleeping at that point. Looking at the code, there are multiple edge cases where that can happen either to the first new message or even during consuming already existing messages.

yunnysunny:
@Tapppi  When are developing a realtime project  with the aim of processing message in low  latency， we can not bear that one message cost such a long time as 500ms.

yunnysunny:
The pull request https://github.com/Blizzard/node-rdkafka/pull/769 has been merged

PR #769, opened 2020-03-22
When the consumer read timeout, try to read after the value of `timeout`

yunnysunny:
The pull request is aimed to resolve the issue of https://github.com/Blizzard/node-rdkafka/issues/709 .

### DIFF
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -17,7 +17,7 @@ var KafkaConsumerStream = require('./kafka-consumer-stream');
 var LibrdKafkaError = require('./error');
 var TopicPartition = require('./topic-partition');
 var shallowCopy = require('./util').shallowCopy;
-var DEFAULT_RETRY_READ_INTERVAL = 500;
+var DEFAULT_CONSUME_LOOP_TIMEOUT_DELAY = 500;
 var DEFAULT_CONSUME_TIME_OUT = 1000;
 util.inherits(KafkaConsumer, Client);
 
@@ -125,8 +125,8 @@ function KafkaConsumer(conf, topicConf) {
   this.globalConfig = conf;
   this.topicConfig = topicConf;
 
-  this._consumeTimeout = DEFAULT_CONSUME_TIME_OUT;
-  this._retryReadInterval = DEFAULT_RETRY_READ_INTERVAL;
+  this._consumeTimeout = DEFAULT_CONSUME_LOOP_TIMEOUT_DELAY;
+  this._consumeLoopTimeoutDelay = DEFAULT_CONSUME_LOOP_TIMEOUT_DELAY;
 }
 
 /**
@@ -134,15 +134,15 @@ function KafkaConsumer(conf, topicConf) {
  * @param {number} timeoutMs - number of milliseconds to wait for a message to be fetched
  */
 KafkaConsumer.prototype.setDefaultConsumeTimeout = function(timeoutMs) {
-  this._consumeTimeout = Number(timeoutMs) || DEFAULT_CONSUME_TIME_OUT;
+  this._consumeTimeout = timeoutMs;
 };
 
 /**
- * Set the default value of waiting for the next time to read when current reading get none data.
- * @param {number} intervalMs - number of milliseconds to sleep before the next reading
+ * Set the default sleep delay for the next consume loop after the previous one has timed out.
+ * @param {number} intervalMs - number of milliseconds to sleep after a message fetch has timed out
  */
-KafkaConsumer.prototype.setDefaultRetryReadInterval = function(intervalMs) {
-  this._retryReadInterval = Number(intervalMs) || DEFAULT_RETRY_READ_INTERVAL;
+KafkaConsumer.prototype.setDefaultConsumeLoopTimeoutDelay = function(intervalMs) {
+  this._consumeLoopTimeoutDelay = intervalMs;
 };
 
 /**
@@ -379,7 +379,7 @@ KafkaConsumer.prototype.unsubscribe = function() {
  * is fetched.
  */
 KafkaConsumer.prototype.consume = function(number, cb) {
-  var timeoutMs = this._consumeTimeout || DEFAULT_CONSUME_TIME_OUT;
+  var timeoutMs = this._consumeTimeout || DEFAULT_CONSUME_LOOP_TIMEOUT_DELAY;
   var self = this;
 
   if ((number && typeof number === 'number') || (number && cb)) {
@@ -419,7 +419,7 @@ KafkaConsumer.prototype.consume = function(number, cb) {
  */
 KafkaConsumer.prototype._consumeLoop = function(timeoutMs, cb) {
   var self = this;
-  var retryReadInterval = this._retryReadInterval;
+  var retryReadInterval = this._consumeLoopTimeoutDelay;
   self._client.consumeLoop(timeoutMs, retryReadInterval, function readCallback(err, message) {
 
     if (err) {
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -17,7 +17,8 @@ var KafkaConsumerStream = require('./kafka-consumer-stream');
 var LibrdKafkaError = require('./error');
 var TopicPartition = require('./topic-partition');
 var shallowCopy = require('./util').shallowCopy;
-
+var DEFAULT_RETRY_READ_INTERVAL = 500;
+var DEFAULT_CONSUME_TIME_OUT = 1000;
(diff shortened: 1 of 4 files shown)


yunnysunny:
You can restart the travis-ci, it can recovery from error.

giulliano-bueno:
This PR seems to be good enough to be merged.

iradul:
I think default value for this timeout is `1000` so this can be a breaking change for some. It would be much better to have a separate timeout field just for handling this scenario.

yunnysunny:
Add a field of `m_retry_read_ms` to set the delay time before next read operation.

lib/kafka-consumer.js
```diff
@@ -17,7 +17,8 @@ var KafkaConsumerStream = require('./kafka-consumer-stream');
 var LibrdKafkaError = require('./error');
 var TopicPartition = require('./topic-partition');
 var shallowCopy = require('./util').shallowCopy;
-
+var DEFAULT_RETRY_READ_INTERVAL = 500;
```
iradul, line 20: Let's call this `CONSUME_LOOP_TIMEOUT_DELAY`

lib/kafka-consumer.js
```diff
@@ -124,15 +125,24 @@ function KafkaConsumer(conf, topicConf) {
...
 KafkaConsumer.prototype.setDefaultConsumeTimeout = function(timeoutMs) {
-  this._consumeTimeout = timeoutMs;
+  this._consumeTimeout = Number(timeoutMs) || DEFAULT_CONSUME_TIME_OUT;
+};
+
+/**
+ * Set the default value of waiting for the next time to read when current reading get none data.
+ * @param {number} intervalMs - number of milliseconds to sleep before the next reading
+ */
+KafkaConsumer.prototype.setDefaultRetryReadInterval = function(intervalMs) {
```
iradul, line 144: Please rename method name to `setDefaultConsumeLoopTimeoutDelay`

lib/kafka-consumer.js
```diff
@@ -124,15 +125,24 @@ function KafkaConsumer(conf, topicConf) {
...
  */
 KafkaConsumer.prototype.setDefaultConsumeTimeout = function(timeoutMs) {
-  this._consumeTimeout = timeoutMs;
+  this._consumeTimeout = Number(timeoutMs) || DEFAULT_CONSUME_TIME_OUT;
+};
+
+/**
+ * Set the default value of waiting for the next time to read when current reading get none data.
+ * @param {number} intervalMs - number of milliseconds to sleep before the next reading
+ */
```
iradul, line 143: Maybe reword like this
```
/**
 * Set the default sleep delay for the next consume loop after the previous one has timed out.
 * @param {number} intervalMs - number of milliseconds to sleep after a message fetch has timed out
 */
```

lib/kafka-consumer.js
```diff
@@ -124,15 +125,24 @@ function KafkaConsumer(conf, topicConf) {
...
+KafkaConsumer.prototype.setDefaultRetryReadInterval = function(intervalMs) {
+  this._retryReadInterval = Number(intervalMs) || DEFAULT_RETRY_READ_INTERVAL;
```
iradul, line 145: Please rename `_retryReadInterval` to `_consumeLoopTimeoutDelay`

src/kafka-consumer.cc
```diff
@@ -1066,14 +1066,23 @@ NAN_METHOD(KafkaConsumer::NodeConsumeLoop) {
   } else {
     timeout_ms = static_cast<int>(maybeTimeout.FromJust());
   }
```
iradul, line 1068: Above these lines, since there are now `3` arguments, validation needs to be fixed:
```
  if (!info[0]->IsNumber()) {
    return Nan::ThrowError("Need to specify a timeout");
  }

  if (!info[1]->IsNumber()) {
    return Nan::ThrowError("Need to specify a sleep delay");
  }

  if (!info[2]->IsFunction()) {
    return Nan::ThrowError("Need to specify a callback");
  }
```

src/workers.cc
```diff
@@ -420,10 +420,12 @@ void KafkaConsumerDisconnect::HandleErrorCallback() {
 
 KafkaConsumerConsumeLoop::KafkaConsumerConsumeLoop(Nan::Callback *callback,
                                      KafkaConsumer* consumer,
-                                     const int & timeout_ms) :
+                                     const int & timeout_ms,
+                                     const int & retry_interval_ms) :
```
iradul, line 424: Please rename to `timeout_sleep_delay_ms`

src/workers.h
```diff
@@ -288,6 +288,7 @@ class KafkaConsumerConsumeLoop : public MessageWorker {
   NodeKafka::KafkaConsumer * consumer;
   const int m_timeout_ms;
   unsigned int m_rand_seed;
+  const int m_retry_read_ms;
```
iradul, line 291: Please rename to `m_timeout_sleep_delay_ms`

lib/kafka-consumer.js
iradul, line 145: Also I think you should not wrap interval within `Number(..)` because validation is done in c++ code

lib/kafka-consumer.js
```diff
@@ -124,15 +125,24 @@ function KafkaConsumer(conf, topicConf) {
...
+  this._retryReadInterval = DEFAULT_RETRY_READ_INTERVAL;
 }
 
 /**
  * Set the default consume timeout provided to c++land
  * @param {number} timeoutMs - number of milliseconds to wait for a message to be fetched
  */
 KafkaConsumer.prototype.setDefaultConsumeTimeout = function(timeoutMs) {
-  this._consumeTimeout = timeoutMs;
+  this._consumeTimeout = Number(timeoutMs) || DEFAULT_CONSUME_TIME_OUT;
```
iradul, line 137: Passing `0` here will not work, also you should not wrap interval within `Number(..)` because:
a) validation is done in c++ code
b) it's kind of breaking change because calling `setDefaultConsumeTimeout('anything but number')` will suddenly start working

Maybe something like `this._consumeTimeout = timeoutMs === 0 ? timeoutMs : timeoutMs || DEFAULT_CONSUME_TIME_OUT;` would be safer.

yunnysunny: Set the value of this._consumeTimeout to timeoutMs directly.

lib/kafka-consumer.js
```diff
@@ -124,7 +125,8 @@ function KafkaConsumer(conf, topicConf) {
   this.globalConfig = conf;
   this.topicConfig = topicConf;
 
-  this._consumeTimeout = 1000;
+  this._consumeTimeout = DEFAULT_CONSUME_LOOP_TIMEOUT_DELAY;
```
iradul, line 128: Here you should use `DEFAULT_CONSUME_TIME_OUT` instead of `DEFAULT_CONSUME_LOOP_TIMEOUT_DELAY`

lib/kafka-consumer.js
```diff
@@ -369,7 +379,7 @@ KafkaConsumer.prototype.unsubscribe = function() {
  * is fetched.
  */
 KafkaConsumer.prototype.consume = function(number, cb) {
-  var timeoutMs = this._consumeTimeout || 1000;
+  var timeoutMs = this._consumeTimeout || DEFAULT_CONSUME_LOOP_TIMEOUT_DELAY;
```
iradul, line 382: Here you should use `DEFAULT_CONSUME_TIME_OUT` instead of `DEFAULT_CONSUME_LOOP_TIMEOUT_DELAY`

src/kafka-consumer.cc
```diff
@@ -1044,16 +1044,16 @@ NAN_METHOD(KafkaConsumer::NodeResume) {
...
+  if (info.Length() < 3) {
     // Just throw an exception
     return Nan::ThrowError("Invalid number of parameters");
   }
 
-  if (!info[0]->IsNumber()) {
-    return Nan::ThrowError("Need to specify a timeout");
+  if (!info[1]->IsNumber()) {
+    return Nan::ThrowError("Need to specify a sleep delay");
   }
```
iradul, line 1054: We need both `info[0]` and `info[1]`:
```
  if (!info[0]->IsNumber()) {
    return Nan::ThrowError("Need to specify a timeout");
  }

  if (!info[1]->IsNumber()) {
    return Nan::ThrowError("Need to specify a sleep delay");
  }
```

iradul:
Thanks!

yunnysunny:
I'm sorry for taking some mistakes. I wrote the code during my overtime work when my head was not clear.

iradul:
No worries, thanks again for PR!

PR #783, opened 2020-04-20
Expose ConsumerStream and ProducerStream interfaces

macabu:
I have a consumer/producer that is implemented on top of these `ConsumerStream` and `ProducerStream` and before this commit (https://github.com/Blizzard/node-rdkafka/commit/66d53834e33c8cb1cfdc1d2d31b5eae851b3841a) they were both being exported.
  
It would be nice to have typings for this interface and not have to use `any`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -123,13 +123,13 @@ export interface WriteStreamOptions extends WritableOptions {
     connectOptions?: any;
 }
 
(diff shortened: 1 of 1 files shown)


NicolasTr:
@iradul Could you release this version to npm? 😄

PR #792, opened 2020-05-04
fix: wrap offset.commit errors with LibrdKafkaError

jpdstan:
These errors weren't properly being wrapped as `LibrdKafkaError`s, and were thus being emitted as [raw error codes](https://github.com/jpdstan/node-rdkafka/blob/c63ef5cd9f5983d572b1d959018d7b2e1ffd34aa/src/callbacks.h#L248) instead of error objects, as we expect in every other error callback on kafka events.

### DIFF
--- a/lib/kafka-consumer.js
+++ b/lib/kafka-consumer.js
@@ -92,11 +92,13 @@ function KafkaConsumer(conf, topicConf) {
 
   if (onOffsetCommit && typeof onOffsetCommit === 'boolean') {
     conf.offset_commit_cb = function(err, offsets) {
+      err = LibrdKafkaError.create(err);
       // Emit the event
       self.emit('offset.commit', err, offsets);
     };
   } else if (onOffsetCommit && typeof onOffsetCommit === 'function') {
     conf.offset_commit_cb = function(err, offsets) {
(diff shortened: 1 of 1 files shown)


lib/kafka-consumer.js
```diff
@@ -92,11 +92,13 @@ function KafkaConsumer(conf, topicConf) {
 
   if (onOffsetCommit && typeof onOffsetCommit === 'boolean') {
     conf.offset_commit_cb = function(err, offsets) {
+      err = LibrdKafkaError.create(err);
```
iradul, line 95: This must to be wrapped into `if (err)` statement

lib/kafka-consumer.js
```diff
@@ -92,11 +92,13 @@ function KafkaConsumer(conf, topicConf) {
...
       // Emit the event
       self.emit('offset.commit', err, offsets);
     };
   } else if (onOffsetCommit && typeof onOffsetCommit === 'function') {
     conf.offset_commit_cb = function(err, offsets) {
+      err = LibrdKafkaError.create(err);
```
iradul, line 101: Same as above

PR #793, opened 2020-05-10
Add ability to write headers to producer stream in object mode

Dishwasha:


iradul:
Thank you for PR.
Please add `headers` at line `235` in `producer-stream.js`:
```js
    try {
      if (Buffer.isBuffer(chunk)) {
        producer.produce(topic, null, chunk, null);
      } else {
        // ↓↓↓↓↓↓ here ↓↓↓↓↓↓
        producer.produce(chunk.topic, chunk.partition, chunk.value, chunk.key, chunk.timestamp, chunk.opaque);
      }
      maybeDone();
    }
```

iradul:
@Dishwasha I'd like to have this feature in the new release which I'm planing to deploy soon. It would be great if you have time to fix this as soon as possible. Thanks

Dishwasha:
My apologies, this completely dropped off my radar.  I've pushed the change you requested.

Issue #731, opened 2019-12-17
Memory leak on disconnect() method usage with producer


Rajat-Sharma-2710:
Hi,

In one of our application we are using nodejs based microservice which internally uses node-rdkafka module to send messages using kafka producer which involves making runtime call scheduled at specific interval to perform below operations:

- Invoking kafka producer connect method 
- Calling producer using kafka produce method to send messages
- Invoking kafka producer disconnect method

Disconnect method usage:
```
let prod =  new Kafka.Producer({
// config properties
})
prod.disconnect()
```
However, we have been noticing memory leak when calling node-rdfafka producer and suspecting that it is disconnect method which is causing the leak. 

Thanks!

codeburke:
What version of Node and node-rdkafka are you using? For clarification, are you creating multiple Producer instances and then `disconnect`ing? Or is there just a few Producer instances instantiated at startup? A simple gist that can isolate and reproduce the problem would be very helpful.

Rajat-Sharma-2710:
@codeburke Yes we are creating multiple producer instances once for each runtime call (scheduled at specific time interval) we make which involves invoking connect() and disconnect() method at every call
```
Node version - 10
node-rdkafka - 2.7.4
```

ottomata:
Hi, I think I'm encountering the same issue in the Kafka Consumer in my use of node-rdkafka in [KafkaSSE](https://github.com/wikimedia/KafkaSSE).  I also connect and then disconnect many Kafka Consumers.

I believe the issue has something to do with `event_cb`.  If I set this to false, the memory leak goes away.  From what I can tell, the native code is holding on to a reference of the _client in its event handler.  

In [client.js](https://github.com/Blizzard/node-rdkafka/blob/master/lib/client.js#L64-L82):
```javascript
  if (!no_event_cb) {
    this._client.onEvent(function eventHandler(eventType, eventData) {
    ...
   }
```

In [connection.cc's NodeOnEvent](https://github.com/Blizzard/node-rdkafka/blob/master/src/connection.cc#L339-L342):
```c++
  Connection* obj = ObjectWrap::Unwrap<Connection>(info.This());

  v8::Local<v8::Function> cb = info[0].As<v8::Function>();
  obj->m_event_cb.dispatcher.AddCallback(cb);
```

[...]

PR #796, opened 2020-05-25
Fix producer/consumer memory leaks #731

iradul:
To dispatch internal events library persists their JavaScript callbacks in the underlying native code. These persistent callbacks disable garbage collection of their holding consumer or producer instance.

One approach to fix this issue is to implement some sort of `destroy()` method that would clean up those persistent callbacks but this doesn't feel natural in JavaScript.

A more natural approach is to wire up and persist those callbacks right before `connect()` call and to clean them up right after `disconnect()` call which is exactly what this PR is all about. This also enables instances re-usability so users can `connect` and `disconnect` as many times as they want and the instance will remain garbage collectible. Callbacks still have to be persisted but this time inside JavaScript (`_cb_configs` object). There are 3 types of those callbacks: global config callback, topic config callback, and producer/consumer specific callbacks.

One downside possibility is that certain events, like maybe `event.log`, could be triggered even before `connect` or after `disconnect`. In my tests this was not the case but again this still could be a possibility.

Ref #731

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -53,32 +53,39 @@ function Client(globalConf, SubClientType, topicConf) {
 
   this._client = new SubClientType(globalConf, topicConf);
 
-  // primitive callbacks from c++ wrapper
-  // need binds on them if we want to broadcast the event via the emitter
+  var extractFunctions = function(obj) {
+    obj = obj || {};
+    var obj2 = {};
+    for (var p in obj) {
+      if (typeof obj[p] === "function") {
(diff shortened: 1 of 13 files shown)


ArneSchulze:
Hi, we tested this fix in our service. Seems to be working fine :+1:

devon-whil:
How can I install this version with yarn/npm? I might be mistaken, but I don't think this has been published yet.

iradul:
@devon-whil it's not published yet but very soon it will be.

PR #816, opened 2020-07-04
Read the download url of librdkafka from Windows environment variable

yunnysunny:
In Windows ,the download of the library of librdkafka may be so slowly, especially when you are in China. It may cost a long time to  download it. Sometime I have to wait for several hours. 
So I add an feature of reading the download url from environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Windows build **is not** compiled from `librdkafka` source but it is rather link
 
 Requirements:
  * [node-gyp for Windows](https://github.com/nodejs/node-gyp#on-windows)  (the easies way to get it: `npm install --global --production windows-build-tools`.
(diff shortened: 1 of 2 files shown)


giulliano-bueno:
LGTM.

yunnysunny:
@iradul Consider this pull request?

iradul:
There is this `BUILD_LIBRDKAFKA` env variable that can be used to compile against system installed version of librdkafka. This doesn't work on Windows but I think it would be great if we could make it work and hence have consistency across platforms.

yunnysunny:
I read the code of binding.gyp, the `BUILD_LIBRDKAFKA` is only an flag to indicate whether to building the source code with the shared library installed in OS already. And my pull request aims at supplying an URL to download nuget package. 
So I think it's two different thing.

yunnysunny:
I  have rename `DOWNLOAD_BASE_URL_LIB_RDKAFKA` to `BUILD_LIBRDKAFKA`.

iradul:
Let's call this variable `NODE_RDKAFKA_NUGET_BASE_URL` and also please leave README file unchanged, I'll update it after I merge this PR.

yunnysunny:
done

yunnysunny:
Expecting this merge can be published to npm. We can not bear the network in China now.

iradul:
@yunnysunny I just released the new version

PR #828, opened 2020-07-29
Improve 'should return ready messages on partition EOF' e2e test

fabianschmitthenner:
Found while working on #803 and wondering why a regression I put there didn't lead to this test failing.
The regression I put there was removing the `if` condition in https://github.com/Blizzard/node-rdkafka/blob/master/src/workers.cc#L537-L539. Then I expected `consume` to directly return, but of cause it still waits for `fetch.max.wait.ms` as we only get EOF notifications at the end of the `fetch` call. Took me some time to figure that out 😂 .

Note that the heuristic of terminating the loop when seeing an `EOF` message is anyway not the best heuristic, really we should wait until we've seen `EOF` messages from _all_ partitions that are consumed, only in cases where we're only assigned to one partition is this the same. But this is beyond this PR.

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -190,10 +190,13 @@ describe('Consumer/Producer', function() {
       var consumeNone = function() {
         // With no new messages, the consume should wait whole timeout
         var start = Date.now();
-        consumer.setDefaultConsumeTimeout(1000);
+        // Set the timeout to 2000ms to see that it actually waits the whole time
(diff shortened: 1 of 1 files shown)


fabianschmitthenner:
:(. PR is still relevant.

fabianschmitthenner:
@iradul pinging you to make sure you see this PR (as I believe it got auto-closed after 3 months), as I'm not able to reopen it myself.

iradul:
Thanks for pointing out

PR #848, opened 2020-10-11
Statically link librdkafka

jjhoughton:
Gets ssl to work with node10 and above on centos 7 by statically
linking librdkafka.

The problem is that centos 7 uses openssl 1.0 where as node10+ is
statically linked to openssl 1.1. As this library first compiles
librafka, which is intern dynamically linked to the version of
openssl that is installed on the OS and not the version included
in node, this can cause instability.

Luckily librdkafka also includes the .a files for statically
linking librdkafka which we can then use to stically link the
library and hense force librdkafka/node-rdkafka to use the version
of openssl included within node.

Signed-off-by: Joshua Houghton <joshua.j.houghton@googlemail.com>


iradul:
Thank you for the PR. This seems like a pretty hacky workaround that would impact all the Linux distros. It would be better if workaround can be applied only to CentOS.

jjhoughton:
I'm not sure it's the right solution myself actually, closing

PR #855, opened 2020-12-07
Fix python 3 support

GaryWilber:
The print line was causing npm install to fail on systems with python 3


deps/windows-install.py
```diff
@@ -14,7 +14,6 @@
 env_dist = os.environ
 downloadBaseUrl = env_dist['NODE_RDKAFKA_NUGET_BASE_URL'] if 'NODE_RDKAFKA_NUGET_BASE_URL' in env_dist else 'https://globalcdn.nuget.org/packages/'
 librdkafkaNugetUrl = downloadBaseUrl + 'librdkafka.redist.{}.nupkg'.format(librdkafkaVersion)
-print 'download librdkafka form ' + librdkafkaNugetUrl
```
iradul, line 17: I think we should better replace it with:
`print('download librdkafka form ' + librdkafkaNugetUrl)`

iradul:
Thanks!

GaryWilber:
> Thanks!

@iradul Could a new release be published with this change? It's a bit tricky to upgrade due to this problem.

PR #856, opened 2020-12-08
Fix tests to be run on travis

fabianschmitthenner:
I realized the tests are not run on travis. The issue appears to be, that the `TRAVIS_BUILD_STAGE_NAME` doesn't capitalize the first letter anymore.

Although the travis documentation claims to do that, I also found some issue about travis not doing that here: https://travis-ci.community/t/travis-build-stage-name-isnt-capitalized/8220.

To work in both cases, let's just enhance the regex to support both.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,14 +13,14 @@ node_js:
 sudo: required
 services: docker
 before_install:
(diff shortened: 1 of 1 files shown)


fabianschmitthenner:
Let's see if it actually works in travis now :).

fabianschmitthenner:
It actually fails in the linting step now 🙈 . Okay, let's make the linting step not fail the travis build for now (as it already improves the process anyway and fixing liniting errors should be done in a subsequent PR).
It shows how important it is to have automatic checks 😄.

PR #857, opened 2020-12-09
Make module self-register (Worker threads fix)

### LINKED ISSUE #831: Module doesn't self-register from a worker thread
**Environment Information**
 - OS [e.g. Mac, Arch, Windows 10]: Alpine (docker)
 - Node Version [e.g. 8.2.1]: node:12-alpine
 - NPM Version [e.g. 5.4.2]: v5
 - C++ Toolchain [e.g. Visual Studio, llvm, g++]:
 - node-rdkafka version [e.g. 2.3.3]: 2.8.1

**Steps to Reproduce**

```
const {
  Worker, isMainThread
} = require('worker_threads');
const kafka = require('node-rdkafka');

if (isMainThread) {
   // Brand new worker
   new Worker(__filename);
}
```

**Additional context**

Even though this issue relates to nodejs internal implementation of workers, nodejs developers wont fix it. However it can be fixed by making native module context-aware. Here is an example for node-lzo: https://github.com/schroffl/node-lzo/pull/11

Twixes:
`node-rdkafka` was incompatible with Worker threads due to legacy context awareness, this fixes the issue for relevant versions of Node.js (>=10.7.0), retaining the old behavior for versions before that. Giving credit where its due, used guidance from https://github.com/nodejs/node/issues/21783#issuecomment-429637117 and work already made in the [node-lzo](https://github.com/schroffl/node-lzo) code base.
Test included.
Resolves #831.

### DIFF
--- a/lib/producer.js
+++ b/lib/producer.js
@@ -7,8 +7,6 @@
  * of the MIT license.  See the LICENSE.txt file for details.
  */
 
-module.exports = Producer;
-
 var Client = require('./client');
 
 var util = require('util');
@@ -289,3 +287,5 @@ Producer.prototype.disconnect = function(timeout, cb) {
     self._disconnect(cb);
   });
 };
+
+module.exports = Producer;
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -73,13 +73,11 @@ void ConstantsInit(v8::Local<v8::Object> exports) {
     Nan::GetFunction(Nan::New<v8::FunctionTemplate>(NodeRdKafkaBuildInFeatures)).ToLocalChecked());  // NOLINT
 }
 
-void Init(v8::Local<v8::Object> exports, v8::Local<v8::Context> context) {
+void Init(v8::Local<v8::Object> exports) {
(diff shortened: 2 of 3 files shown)


iradul:
Thanks for the PR.
I've run a few local tests after merging this and I've got occasional crashes. It seems we may need to do more work on this to make the library reliably runs within worker threads.

Twixes:
Yeah, us at PostHog too. Had some awful segfaults. We've since switched to https://kafka.js.org/ and it's a much smoother experience, though probably somewhat less performant than tapping into C.

yunnysunny:
I have try change the addon library from Nan to node-addon-api to support Worker threads. But I'm not familiar with C++, it also have some bugs. The project link: https://github.com/node-rdkafka-next/node-rdkafka-next

yunnysunny:
I have try change the addon library from Nan to node-addon-api to support Worker threads. But I'm not familiar with C++, it also have some bugs. The project link: https://github.com/node-rdkafka-next/node-rdkafka-next

Twixes:
I've switched to `NAN_MODULE_WORKER_ENABLED` here, which should be exactly the right (and official) macro for this. All tests green, but some stability tests need to be done here to ensure there are no crashes.

Twixes:
What does this solution look like to you?

PR #866, opened 2021-01-13
Reduce event listener type instantiation depth

josh-feldman:
The use of chained `extends` statements in the `EventListener` type generic results in a very deep type instantiation for the listener function.

If the listener callback also utilizes generics, Typescript will complain that `Type instantiation is excessively deep and possibly infinite.ts(2589)`

The depth that the `EventListener` generic is taking up can be reduced to a single level by using a type map with the `keyof` keyword.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -141,34 +141,36 @@ type KafkaClientEvents = 'disconnected' | 'ready' | 'connection.failure' | 'even
 type KafkaConsumerEvents = 'data' | 'partition.eof' | 'rebalance' | 'rebalance.error' | 'subscribed' | 'unsubscribed' | 'unsubscribe' | 'offset.commit' | KafkaClientEvents;
 type KafkaProducerEvents = 'delivery-report' | KafkaClientEvents;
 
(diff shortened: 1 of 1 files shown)


iradul:
Thanks, this is great.

index.d.ts
```diff
@@ -141,34 +141,36 @@ type KafkaClientEvents = 'disconnected' | 'ready' | 'connection.failure' | 'even
...
+    // event messages 
+    'event.error': (error: LibrdKafkaError) => void,
+    'event.stats': (eventData: any) => void,
+    'event.log': (eventData: any) => void,
+    'event.event': (eventData: any) => void,
+    'event.throttle': (eventData: any) => void,
+    // ### Consumer only
+    // domain events
+    'data': (arg: Message) => void,
+    'partition.eof': (arg: Message) => void,
```
iradul, line 159: This should be `'partition.eof': (arg: EofEvent) => void`

josh-feldman: Good catch, thanks. Resolved in 097417e.

PR #879, opened 2021-03-24
Update librdkafka from 1.5.3 -> 1.6.1

fabianschmitthenner:
I tried to look at some recent similiar procedure how this is done (like in https://github.com/Blizzard/node-rdkafka/commit/6810bdb9369e331170d248f5acc5cfb7517f4891#).

I executed tests (all succeeded) and e2e tests. They always succeed the second time, but the Consumer/Producer, "should be able to produce, consume messages, read position: subscribe/consumeOnce" usually fails the first time (when the topic  doesn't exist yet), because the consumer emits a "Broker: Unknown topic or partition" error (error code 3) which leads to a `length` can't be called on `undefined` error. I think this is a regression already introduced by the update to librdkafka 1.5.0, so not a regression here. As the update description from 1.5.0 says:

> Subscribing to non-existent and unauthorized topics will now propagate
errors RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART and
RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED to the application through
the standard consumer error (the err field in the message object).

~~I also haven't figured out how to generate `config.d.ts`, `errors.d.ts` and `lib/error.js`, so I didn't change them so far.~~

~~`config.d.ts` has a comment, that this is created by a derivative work of https://gist.github.com/ntgn81/066c2c8ec5b4238f85d1e9168a04e3fb, but it doesn't say which derivative work.~~

~~@iradul Any pointers how to generate these files?~~

~~I think it would also be helpful if the scripts to generate these files would also live in this repository.~~

The generation script is indeed part of the repository, as explained by @iradul.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ I am looking for *your* help to make this project even better! If you're interes
 
 The `node-rdkafka` library is a high-performance NodeJS client for [Apache Kafka](http://kafka.apache.org/) that wraps the native  [librdkafka](https://github.com/edenhill/librdkafka) library.  All the complexity of balancing writes across partitions and managing (possibly ever-changing) brokers should be encapsulated in the library.
 
-__This library currently uses `librdkafka` version `1.5.3`.__
+__This library currently uses `librdkafka` version `1.6.1`.__
 
 ## Reference Docs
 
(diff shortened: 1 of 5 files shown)


iradul:
Just run `node ci/librdkafka-defs-generator.js`. Also worth mentioning this script is part of the deployment process so even if it's not executed manually it will be before the lib gets published to NPM.

fabianschmitthenner:
okay, thanks 👍 . Directly executed it :)

iradul:
Looks good, thank you!

GaryWilber:
@iradul Could a new version be published to npm with this change? I would like to start using the new librdkafka version 😃

PR #883, opened 2021-04-16
Update README

raunc:
Changes:
* Be more explicit about possible consequences of using `consumer.consume(cb)`


iradul:
Good point. This is also true for both `consumer.consume()` and `consumer.consume(number, cb)` (topics with slow message rate), basically any kind of consumption.

I think it would be even better to mention this as a general rule of thumb for use with multiple consumers.

PR #909, opened 2021-08-19
Specify absolute path to win action in binding.gyp and minor fixes to windows-install.py

lkallas:
Refer to https://github.com/nodejs/node-gyp/issues/2485 why absolute path is needed when using latest `node-gyp`.


lkallas:
The issue was in the `gyp-next` package embedded in `node-gyp` which is now fixed and released with `node-gyp@8.2.0`
This fix is not needed when latest `node-gyp` is installed. Closing PR.

PR #916, opened 2021-09-26
provide TopicPartitionOffset to HighLevelProducer callback

tgerk:
I was disappointed that I couldn't get all three of topic, partition, and offset from Delivery Report in my HighLevelProducer callback function, the offset alone seemed somehow incomplete.

I hope someone may find this extension useful.


tgerk:
I rebased my change, hope you will consider it.

PR #935, opened 2022-01-28
Add support for Buffer types in Kafka Headers

CharlesSwiftConnect:
The node-rdkafka API accepts Buffers or strings in KafkaHeaders (in line with the librdkafka API), but was converting all Header Values to strings. You can test this by attempting to send a Header value such a Buffer generated from a Uint8 array with values above 127 or including 0s. The first 0 encountered will be treated as a null termination, and bytes above 127 will be spread across multiple array entries, changing the length and values of the uint8 array.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -32,7 +32,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'cpp', 'javascript', 'python' ]
+        #  'cpp', 
+        language: ['javascript', 'python' ]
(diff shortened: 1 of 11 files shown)


loay:
@iradul would it be possible to merge this one? it fixes an issue that we are having. Thanks.

loay (approved):
Looks good to me

iradul:
Thank you for the PR.
The code changes should be broken up into a few PRs:
* upgrade to `librdkafka` `1.8.2`, however, that is already done
* fix the Windows install (changes in `deps/windows-install.py`)
* Buffer headers implementation (this PR)

PR #945, opened 2022-03-18
Docs: fix alpine example due to python version

cbrwflo:
Fixes the Alpine example, which [removed](https://git.alpinelinux.org/aports/commit/?h=3.12-stable&id=5ad0ec7da1064361cc74d56edf7524960f49ef9b) the generic `python` package in v3.12.

node-gyp supports python3 since v5.x:
https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md


iradul:
Thank you for the PR. Could you also update the image from `FROM node:8-alpine` to `FROM node:14-alpine`?

cbrwflo:
Sorry for the delay -- tested and this builds fine.

Issue #953, opened 2022-05-04
memory leak on exception in produce() when using opaque data 


emelski:
**Environment Information**
 - OS: Linux
 - Node Version: 14
 - NPM Version: 6.14.16
 - C++ Toolchain: g++
 - node-rdkafka version: 2.10.1

**Steps to Reproduce**

`producer.produce()` allows the caller to provide arbitrary data to be included in delivery reports via the `opaque` parameter.  Normally the opaque data is persisted by the call to `produce()` and then released when the delivery report is made.  In the case of an error in `produce()`, no delivery report ever comes for the message, so the opaque data is never released, resulting in a memory leak.

Assuming a kafka instance running at localhost:9093 (I used the bitnami docker image in my test), and that kafka is setup _not_ to auto-create unknown topics, this script makes it possible to see the leak.

```
const {once} = require('events')
const Kafka = require('node-rdkafka')

function dr(err, report) {
  console.log({err}, 'delivery report')
}

async function main() {
  const prod = new Kafka.Producer({
    'dr_cb': true,
    'metadata.broker.list': 'localhost:9093',
    'security.protocol': 'sasl_plaintext',
    'sasl.mechanism': 'PLAIN',
    'sasl.username': 'user',
    'sasl.password': 'bitnami',
    'topic.metadata.propagation.max.ms': 500
  }, {})
  prod.on('delivery-report', dr)
  console.log('connecting')
  prod.connect()
  await once(prod, 'ready')
  console.log('connected')

  for (let i = 0 ; i < 1000 ; ++i) {
    console.log(process.memoryUsage(), `producing ${i}`)
    try {
      let opaque = { leaked: Buffer.from(`${i}`.repeat(1048576)) }
      prod.produce('bogus', -1, Buffer.from('my message'), null, Date.now(), opaque)
      prod.flush(500, () => { })
    } catch (err) {
      console.log({err}, `error producing ${i}`)
    }
    await new Promise(r => setTimeout(r, 100));
  }
}

main()
```

emelski:
Submitted a PR that fixes the leak and which I've been using in production for several months.

https://github.com/Blizzard/node-rdkafka/pull/954

PR #954, opened 2022-05-04
fix: release opaque data on error in produce

emelski:
When there is an error trying to produce() messages,
no delivery report is ever generated for the message.
If the user specified `opaque` data for the message,
the call to produce() will have "pinned" it, but with
no delivery report, it is never released.  This change
checks for this condition and releases the opaque
data when this happens.

Fixes: #953


iradul:
Thank you for the PR!

PR #968, opened 2022-08-01
Add support for Buffer types in Kafka Headers

CharlesSwiftConnect:
The node-rdkafka API accepts Buffers or strings in KafkaHeaders (in line with the librdkafka API), but was converting all Header Values to strings. You can test this by attempting to send a Header value such a Buffer generated from a Uint8 array with values above 127 or including 0s. The first 0 encountered will be treated as a null termination, and bytes above 127 will be spread across multiple array entries, changing the length and values of the uint8 array.


CharlesSwiftConnect:
created cleaner PR to replace PR https://github.com/Blizzard/node-rdkafka/pull/935 (same commit)

esimphlet:
Sound great !

tleasure:
May I ask when this will be merged? I just ran into the issue and confirmed the patch from @CharlesSwiftConnect  allows my buffers to go through correctly.

PR #970, opened 2022-08-19
Fix bug that header values are not null-terminated

jan-lindemann:
There are currently two bugs in the producer code when dealing with buffers as header values.

The first issue is that the data of buffers passed as header values is cut off on the first null byte. According to the documentation (see https://kafka.apache.org/documentation/#recordheader) headers of Kafka messages can be arbitrary byte sequences. The bug is caused by the use of the copy constructor in `std::string`, which stops copying on the null byte. Using the copy constructor where the length of the byte sequence is passed explicitly will fix this issue.

A second issue is that a buffer is treated similar to a string and that its content is passed to the UTF8String constructor. This will modify the byte sequence, as this constructor will transform the passed data from the node internal UTF-16 string representation into a UTF-8 representation. Since buffers can be arbitrarily encoded data and not necessarily strings, buffers should not be passed to the UTF8String constructor. Instead, their value should be used as is.

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -409,6 +409,27 @@ describe('Consumer/Producer', function() {
     ];
     run_headers_test(done, headers);
   });
+
+  it('should be able to produce and consume messages with one header with NULL byte', function(done) {
+    var b = Buffer.alloc(3);
+    b.writeUInt8(1, 0);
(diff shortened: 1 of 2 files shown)


iradul:
Thank you for the PR.
Fixing the first issue implies a fix for the second one since utf-16 should be treated as Buffer.

PR #988, opened 2022-12-18
Upgrade NAN, bindings, and deprecate node versions too low

webmakersteve:


GaryWilber:
I made these changes as part of #997

PR #991, opened 2022-12-22
Upgrade Mocha to 10

SpaceFozzy:
When installing the NPM dependencies I see `4 vulnerabilities (1 moderate, 2 high, 1 critical)`. This is just because the Mocha version is outdated.

This PR updates Mocha from version 5 to version 10. The only breaking change is updating [the configuration file](https://mochajs.org/#configuring-mocha-nodejs) that defines the `ui` as `export`. Since mocha now requires that the config file goes in the project root, or that you pass a custom path as a param, I added it to the`package.json` for simplicity.

Changes:
 * Upgrade mocha from 5 to 10 and add with the new `npm-shrinkwrap.json`
 * Remove the old `mocha.opts` used to configure Mocha
 * Update the `package.json` with a `mocha` property containing the Mocha configuration

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -17,7 +17,7 @@
         "bluebird": "^3.5.3",
         "jsdoc": "^3.4.0",
         "jshint": "^2.10.1",
-        "mocha": "^5.2.0",
+        "mocha": "^10.2.0",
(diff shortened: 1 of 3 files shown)


GaryWilber:
Thanks for the PR. I updated the package as part of #997

Issue #940, opened 2022-02-22
node_modules size more then 100mo


selmi38:
Hello,
the size of node-rdkafka is about  167.9 Mo inside node_modules, i looked into it and i found that 

node_modules/node-rdkafka/deps  = 135.4 Mo
node_modules/node-rdkafka/build/deps/librdkafka.a = 15.1 mo
node_modules/node-rdkafka/build/deps/librdkafka.a = 5.7 mo

seems like removing them have no impact, everything work fine.
its possible please to make sure to delete them

tatianajiselle:
if you are containerizing your project, you can manually delete these libraries in your dockerfile after a npm install:

```
RUN rm -rf /full/path

```

Elsvent:
Per investigate
`
bash-5.0# du -h -d 1
11.3M	./src-cpp
212.0K	./win32
59.3M	./src
76.0K	./debian
2.1M	./tests
1.7M	./packaging
160.1M	./examples
148.0K	./mklove
235.6M	.
bash-5.0# pwd
/node_modules/node-rdkafka/deps/librdkafka
`
example take 160 M
Just a reference for you

raymcdermott:
My experience today:

```shell
$ du -sm node-rdkafka # -m reports size in Mb
432     node-rdkafka
$ cd node-rdkafka
du -sm * | sort -rn       
366     deps
66      build # <--------------------------- Actually used by nodejs
```

Removing build stopped the runtime from working.

Removing deps has no effect on the runtime. 

Would be great to clean up after the build in your package.json scripts

akaustel:
The issue is not really the package size, but the build script does not clean up after building on installation. It seems the `.node` library actually used is ~0.5MB, while unnecessary build deps are ~470MB.

```
# du -hd1 node_modules/node-rdkafka/
72K	node_modules/node-rdkafka/ci
136K	node_modules/node-rdkafka/e2e
52K	node_modules/node-rdkafka/util
244K	node_modules/node-rdkafka/lib
469M	node_modules/node-rdkafka/deps
71M	node_modules/node-rdkafka/build
444K	node_modules/node-rdkafka/src
100K	node_modules/node-rdkafka/examples
76K	node_modules/node-rdkafka/bench
212K	node_modules/node-rdkafka/test
542M	node_modules/node-rdkafka/
```

Would it make sense to clean this?

PR #1034, opened 2023-08-08
Add examples to npmignore

neil-b-patel:
One-liner to address issue https://github.com/Blizzard/node-rdkafka/issues/940

The `deps/librdkafka/examples` directory shouldn't be included in published artifacts, especially with how heavy it can make the package. It's also likely that most users will end up navigating to this repository for examples and usage anyways

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -8,3 +8,4 @@ deps/librdkafka/config.h
 build
 .github
 .vscode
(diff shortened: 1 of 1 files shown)


neil-b-patel:
@GaryWilber saw that you been the last few commits to `master`, can you take a look at this PR?

PR #1061, opened 2023-12-06
feat: add global consume timeout

bojan-rajkovic-simplisafe:
This solves a similar problem to #1052, but does it with an approach that I think is a little cleaner and doesn't change existing consumer behavior — even if the existing consumer behavior is arguably wrong, all behavioral changes are susceptible to the [spacebar heating problem](https://xkcd.com/1172/).

Rather than making the specified timeout both the single-message consume timeout *and* the batch timeout, it introduces a new value called the total consume timeout, which controls how long we can spend trying to fill a batch of `n` messages with the `consume(num, cb)` variant, and checks it on every iteration of the consume loop.

I did borrow the tests from that PR though, as they provided a pretty elegant framework to check the approach.

### DIFF
--- a/e2e/both.spec.js
+++ b/e2e/both.spec.js
@@ -359,7 +359,7 @@ describe('Consumer/Producer', function() {
         // produced over 990 seconds, and then a 13th is produced and consumed
         // in the "final" loop where elapsed time > 10,000.
         t(messages.length == 13, 'Batch should have consumed 13 messages, instead consumed ' + messages.length + ' messages');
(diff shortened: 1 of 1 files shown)


bojan-rajkovic-simplisafe:
Fixed — in the process, ended up cleaning up C++-land a bit and killing some dead code + some code that I'm not sure I understand how it ever worked. :)

bojan-rajkovic-simplisafe:
@GaryWilber @iradul It might be good to get one of you to look at both this and #1053 and decide which approach to take, merge one, and do some testing before cutting a release. I'm fairly confident in mine, but it could probably use more testing regardless…

PR #1065, opened 2024-01-11
Feat: consume messages per partition

martijnimhoff:
Pull request checklist:

- [x] Include new test cases (either end-to-end or unit tests) with your change.
  > Existing test cases do not include consuming via the standard API. So did not make those for this case.
- [x] Follow our style guides.
- [x] Make sure all tests are still passing and the linter does not report any issues.
  > Current master has only quite a few linter errors. I made sure to not add any new errors and I've fixed a few.
- [x] End files with a new line.
- [x] Document the new code in the comments (if it is JavaScript) so the documentation generator can update the reference documentation.
- [x] Avoid platform-dependent code.
- [x] Note: If making modifications to the underlying C++, please use built-in precompiler directives to detect such platform specificities. Use Nan whenever possible to abstract node/v8 version incompatibility.
- [x] Make sure your branch is up to date and rebased.
- [x] Squash extraneous commits unless their history truly adds value to the library.


martijnimhoff:
We will first review this internally at withthegrid.

PR #1067, opened 2024-03-05
Fix memory leak

martijnimhoff:


martijnimhoff:
Sorry, should have been a PR on our own fork.

PR #1069, opened 2024-03-08
Fix outdated package-lock.json

martijnimhoff:
The package-lock file contains the wrong package version. A `npm install` will fix that, but also cause some other changes. AFAIK this does not harm functionality, it just makes the lock file smaller.


GaryWilber:
The latest release now has an updated package-lock on lock version 3. Closing this PR

PR #1074, opened 2024-04-08
Update node-gyp to fix build error in Python 3.12

alexzel:
Running `npm install` throws node-gyp error `ModuleNotFoundError: No module named 'distutils'` in Python `3.12` (the default version on MacOS 12). The issue was fixed in the latest `node-gyp` releases. Please see https://github.com/nodejs/node-gyp/issues/2869 Thank you!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "node-rdkafka",
-  "version": "v2.17.0",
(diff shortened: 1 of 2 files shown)


GaryWilber:
Thanks for the PR. I included this in #1049

PR #1075, opened 2024-04-16
Add setToken API for OAuthBearer authentication flow

andrewstanovsky:
This change adds new setToken API based on librdkafka's one to allow node-rdkafka clients authenticate using OAuthBearer flow. SetToken accepts token string only, no expiry is expected as it's always set to maximum value with librdkafka at C++ level. The client code is supposed to call setToken again to refresh a token when needed. The expiry is used in refresh callback flow of librdkafka which is not used in this change to avoid complexity.


examples/oauthbearer-default-flow.md
```diff
@@ -0,0 +1,80 @@
+Producer, Consumer and HighLevelProducer:
```
mpodwysocki, line 1: Why is this just a markdown file instead of something executable?  And is it necessary to put the copyright header on all code snippets?

PR #1100, opened 2024-11-06
fix: `util.isError` API is deprecated

martijnimhoff:
When node 22 is used you get the following deprecation warning:

```
(node:3751956) [DEP0048] DeprecationWarning: The `util.isError` API is deprecated. Please use `ObjectPrototypeToString(e) === "[object Error]" || e instanceof Error` instead.
```

This fixes that and makes this repo compatible with node 22. No other changes are needed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


GaryWilber:
This was done in #1119

PR #1104, opened 2024-11-18
Add node 22 and 23 to test matrix

GaryWilber:


milindl:
[v8 removed v8::CopyablePersistentTraits ](https://chromium.googlesource.com/v8/v8/+/c51a81e705d2e407052c6c281214d9d7e209cbab)from the version of v8 used in node v23, and using Nan::CopyablePersistentTraits doesn't seem to work, it gives a template substitution error for node v18. [We're trying to get around it ](https://github.com/confluentinc/confluent-kafka-javascript/pull/158/commits/e911b62b669febc62bc3313191298f3930cf0996)by replacing `std::vector<v8::Persistent<v8::Function, Nan::CopyablePersistentTraits<v8::Function> > > callbacks` with `std::vector<Nan::Callback*> callbacks;`.

PR #1110, opened 2024-12-20
Fix worker_threads support

Qard:
I'm working on making rdkafka work with worker_threads. There's a few issues with how it works presently which I've _mostly_ resolved, but a few issues still need fixing.

Changes required to support worker_threads:

- No globals, shared data needs to live in PerIsolateData.
- No using `uv_default_loop()`, need to get the current loop for the thread.
- Many long-lived things need to use node cleanup hooks to ensure they are deleted when the thread is torn down.

The first two are fully resolved, but I think there's so issues yet with the thread teardown cleanup. This will need a bit more work yet so marking as draft...


src/common.h
```diff
@@ -12,7 +12,6 @@
 
 #include <nan.h>
 
-#include <iostream>
 #include <string>
```
rozzilla, line 15:
```suggestion
#include <iostream>
#include <string>
```
Hi @Qard 👋🏼  The import should be reverted. The code doesn't compile ATM

rozzilla:
```cc
../src/common.cc:18:8: error: no member named 'cerr' in namespace 'std'
   18 |   std::cerr << "% " << str.c_str() << std::endl;
      |   ~~~~~^
```

Qard: Odd. Seemed to work for me on macOS and it complained that it was unused. I've restored it.

mcollina:
@GaryWilber can you take a look?

Issue #1057, opened 2023-11-29
Segfault when disconnecting right after consuming


abarisain:
**Environment Information**
 - OS [e.g. Mac, Arch, Windows 10]: macOS 14, M2 Pro
 - Node Version [e.g. 8.2.1]: 20, reproducible on 18 too
 - NPM Version [e.g. 5.4.2]: 10.2.3
 - C++ Toolchain [e.g. Visual Studio, llvm, g++]: llvm
 - node-rdkafka version [e.g. 2.3.3]: 2.8.0

**Steps to Reproduce**

Hello,

I'm writing automated tests with playwright and need to access kafka from them. This implies connecting and disconnecting many times from multiple brokers during the life of a process.

My consumer wrapper returns a promise that resolves once a message has been consumed on any partition. Before resolving, I disconnect from the brokers as the connection will not be reused.

When I try consuming multiple topics from multiple brokers at once and disconnect, I experience crashes:
```
PID 48415 received SIGSEGV for address: 0x0
0   segfault-handler.node               0x00000001084a0d48 _ZL16segfault_handleriP9__siginfoPv + 296
1   libsystem_platform.dylib            0x0000000181b9fa24 _sigtramp + 56
2   node-librdkafka.node                0x000000010850da0c _ZNK3Nan8Callback5Call_EPN2v87IsolateENS1_5LocalINS1_6ObjectEEEiPNS4_INS1_5ValueEEEPNS_13AsyncResourceE + 64
3   node-librdkafka.node                0x000000010850a05c _ZNK3Nan8Callback4CallEiPN2v85LocalINS1_5ValueEEE + 192
4   node-librdkafka.node                0x0000000108523ee8 _ZN9NodeKafka7Workers24KafkaConsumerConsumeLoop21HandleMessageCallbackEPN7RdKafka7MessageENS2_9ErrorCodeE + 700
5   node-librdkafka.node                0x000000010852679c _ZN9NodeKafka7Workers13MessageWorker11WorkMessageEv + 112
6   node                                0x000000010353c5b4 uv__async_io + 268
7   node                                0x000000010354e68c uv__io_poll + 1020
8   node                                0x000000010353cb78 uv_run + 476
9   node                                0x0000000102a9d754 _ZN4node21SpinEventLoopInternalEPNS_11EnvironmentE + 256
[...]

PR #1136, opened 2025-08-08
Fix: segfault due to use after free in workers thread

constantind:
When NodeDisconnect calls worker.WorkComplete() and Destroy() sets the callback to null, while the thread is stopping or starting in a race, creating "use after free" segmentation fault in the message callback handler on Linux.

Closes: https://github.com/Blizzard/node-rdkafka/issues/1057

cc: @GaryWilber


constantind:
Yes, it is better in the worker itself, i will run a test and update

PR #1148, opened 2026-08-03
Support authenticated NuGet downloads on Windows

sathyarajagopal:
## Summary

- allow Windows NuGet downloads to include configurable HTTP headers
- support Basic, Bearer, API-key, access-token, and custom authentication
- document `NODE_RDKAFKA_NUGET_HEADERS`
- add unit tests for authenticated and unauthenticated requests

## Testing

- `python3 deps/windows-install.spec.py` (4 tests passed)
- `python3 -m py_compile deps/windows-install.py deps/windows-install.spec.py`
- `git diff --check`

The full project test and lint targets could not be run locally because project dependencies are not installed and the existing Makefile requires Python 2.


sathyarajagopal:
@GaryWilber @edenhill @webmakersteve Could you please review this and help push this through? 

For context, our company policy blocks external downloads, so all artifacts must be retrieved through our internal JFrog Artifactory. Because this internal repository requires authentication, unauthenticated requests are rejected.
When running "npm install node-rdkafka" with a custom NODE_RDKAFKA_NUGET_BASE_URL, the librdkafka.redist download failed because the current urllib.urlopen call does not include authentication headers. This PR adds support for configurable request headers, enabling downloads from authenticated private repositories.

